PR #1, opened 2018-10-09
Fusion for Free

robrix:
This PR implements the ideas of _Fusion for Free_, replacing the free monad definition of `Eff` with a codensity monad definition, and interpreting all effectful computations via term algebras and monads.

This also reduces the requirements for carriers to a minimum: `Carrier` and `TermAlgebra` instances, no functor/applicative/monad instances necessary. I don’t yet have a great handle on the interaction between higher-order effects and fused handlers, except that it seems like suspension/resumption is still necessary in order to implement e.g. `Catch`.

I’ve 🔥’d `Resumable`, as it’s no longer possible to close over an exception handler; instead, resumable exceptions should be implemented as individual error types with distinct handlers for each use of `runResumableWith`. This is _slightly_ less convenient, but should be no less expressive in practice.


higher-order-effects.cabal
```diff
@@ -13,23 +13,13 @@ extra-source-files:  ChangeLog.md
...
 library
-  exposed-modules:     Control.Carrier
-                     , Control.Carrier.Either
-                     , Control.Carrier.Identity
-                     , Control.Carrier.List
-                     , Control.Carrier.Maybe
-                     , Control.Carrier.Reader
-                     , Control.Carrier.Split
-                     , Control.Carrier.State
-                     , Control.Carrier.Writer
```
robrix: Carriers are now one-to-one with effect handlers, meaning that we can’t e.g. have two different `TermAlgebra` instances for `StateH`. Thus, we 🔥 the separate carrier modules and instead define the carriers in the effect files alongside the `run*` functions.

src/Control/Effect.hs
```diff
@@ -25,20 +17,34 @@ module Control.Effect
...
-import Control.Monad (ap, liftM)
+import Control.Monad (liftM, ap)
 import Control.Monad.Fail
 import Control.Monad.IO.Class
 import Prelude hiding (fail)
 
-data Eff effects a
-  = Return a
-  | Eff (effects (Eff effects) (Eff effects a))
+newtype Eff h a = Eff { unEff :: forall x . (a -> h x) -> h x }
```
robrix: We no longer need a free-like `Eff` monad at all; instead, `Eff` is redefined as the codensity monad.

src/Control/Effect.hs
```diff
@@ -48,114 +54,50 @@ class Effect sig where
...
-  --   For higher-order effects (which do contain higher-order positions), the 'Carrier' actions @c n@ must be 'resume'd to obtain the necessary @n@ action, and the state passed along to the continuation via 'resume' and 'wrap'.
-  handle :: (Carrier c f, Monad n)
+  -- | Handle any effects in higher-order positions by threading the carrier’s state all the way through to the continuation.
+  handle :: (Functor f, Monad n)
          => f ()
-         -> sig (c n) (c n a)
-         -> sig n (c n a)
+         -> (forall x . f (m x) -> n (f x))
+         -> sig m (m a)
+         -> sig n (n (f a))
```
robrix: This is the bog-standard handler from _Effect Handlers in Scope_.

src/Control/Effect.hs
```diff
@@ -48,114 +54,50 @@ class Effect sig where
...
+
+
+class Effect sig => TermAlgebra h sig | h -> sig where
+  var :: a -> h a
+  con :: sig h (h a) -> h a
```
robrix: `var` and `con` must be used to construct effectful computations for correctness’ sake. Fortunately this is extremely easy to abide by since `send` &c. use these.

src/Control/Effect/Error.hs
```diff
@@ -15,17 +14,23 @@ instance Effect (Error exc) where
...
-  where alg (Throw e)     = EitherH (pure (Left e))
-        alg (Catch m h k) = EitherH (runEitherH m >>= runEitherH . either (k <=< h) k)
+runError :: TermMonad m sig => Eff (ErrorH exc m) a -> m (Either exc a)
+runError = runErrorH . runEff var
+
+newtype ErrorH e m a = ErrorH { runErrorH :: m (Either e a) }
+
+instance TermMonad m sig => TermAlgebra (ErrorH e m) (Error e :+: sig) where
+  var a = ErrorH (pure (Right a))
+  con = alg \/ (ErrorH . con . handle (Right ()) (either (pure . Left) runErrorH))
```
robrix: This is really quite similar to the presentation we currently use in the `effects` package.

PR #15, opened 2018-10-12
Resumable errors

robrix:
This PR defines a `Resumable` effect type and handlers, and an existentially-quantified `SomeError` type suitable for wrapping type-indexed errors.

- [x] Defines a `Resumable` effect.
- [x] Defines a `runResumable` handler.
- [x] Defines a `SomeError` hiding its error type’s index.
- [x] Defines a `runResumableWith` handler parameterized by a higher-order effect handler.

### DIFF
--- a/src/Control/Effect/Resumable.hs
+++ b/src/Control/Effect/Resumable.hs
@@ -1,1 +1,5 @@
+{-# LANGUAGE ExistentialQuantification #-}
 module Control.Effect.Resumable where
+
+data Resumable exc k
+  = forall a . Resumable (exc a) (a -> k)
--- a/src/Control/Effect/Resumable.hs
+++ b/src/Control/Effect/Resumable.hs
@@ -5,6 +5,7 @@ module Control.Effect.Resumable
 , SomeError(..)
 , runResumable
 , ResumableH(..)
+, runResumableWith
 ) where
 
 import Control.Effect.Handler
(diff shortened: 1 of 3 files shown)


src/Control/Effect/Resumable.hs
```diff
@@ -0,0 +1,115 @@
...
+runResumableWith :: (Carrier sig m, Monad m)
+                 => (forall x . err x -> m x)
+                 -> Eff (ResumableWithH err m) a
+                 -> m a
+runResumableWith with = runResumableWithH with . interpret
+
+newtype ResumableWithH err m a = ResumableWithH ((forall x . err x -> m x) -> m a)
+
+runResumableWithH :: (forall x . err x -> m x) -> ResumableWithH err m a -> m a
+runResumableWithH f (ResumableWithH m) = m f
```
robrix: The `ResumableWithH` carrier takes a rank-_n_ higher-order function, and `flip` doesn’t play nicely with rank-_n_ constructors. Therefore, instead of using a record selector and flipping it, we just define the accessor as a function with the arguments in the order we want.

src/Control/Effect/Resumable.hs
```diff
@@ -0,0 +1,115 @@
...
+--   prop> (SomeError (Const a) == SomeError (Const b)) == (a == b)
+instance Eq1 err => Eq (SomeError err) where
+  SomeError exc1 == SomeError exc2 = liftEq (const (const True)) exc1 exc2
+
+-- | Ordering for 'SomeError' is determined by an 'Ord1' instance for the error type.
+--
+--   Note that since we can’t tell whether the type indices are equal, let alone what 'Ord' instance to use for them, the comparator passed to 'liftCompare' always returns 'EQ'. Thus, 'SomeError' is best used with type-indexed GADTs for the error type.
+--
+--   prop> (SomeError (Identity a) `compare` SomeError (Identity b)) == EQ
+--   prop> (SomeError (Const a) `compare` SomeError (Const b)) == (a `compare` b)
```
robrix: I’m pretty happy with how these doctests turned out.

PR #39, opened 2018-10-28
Random

robrix:
This PR defines a `Random` effect, with `MonadRandom` and `MonadInterleave` instances for lifting `RandomGen` generators into `Eff` computations.


src/Control/Effect/Random/Internal.hs
```diff
@@ -0,0 +1,25 @@
...
+( Random(..)
+) where
+
+import Control.Effect.Carrier
+import qualified System.Random as R (Random(..))
+
+data Random m k
+  = forall a . R.Random a => Random (a -> k)
+  | forall a . R.Random a => RandomR (a, a) (a -> k)
+  | forall a . Interleave (m a) (a -> k)
```
robrix: Really pleased with how this turned out. Not all `RandomGen` instances have a viable `split`, but for those that do, we can trivially implement `MonadInterleave`.

PR #47, opened 2018-10-28
Handler ergonomics

robrix:
This PR defines a `handleSum` function which has better ergonomics than `\/`: not only does it have a pronounceable name, it has better error messaging when using typed holes than the erstwhile standard pattern of an algebra defined in the where clause.

- [x] Do we really want to keep `\/` around now?
- [x] Update the README to use `handleSum`.
- [x] :memo: `handleSum`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,9 +195,9 @@ newtype TeletypeIOC m a = TeletypeIOC { runTeletypeIOC :: m a }
 instance (Carrier sig m, MonadIO m) => Carrier (Teletype :+: sig) (TeletypeIOC m) where
   ret = TeletypeIOC . ret
 
-  eff = TeletypeIOC . (alg \/ eff . handleCoercible)
-    where alg (Read    k) = liftIO getLine      >>= runTeletypeIOC . k
-          alg (Write s k) = liftIO (putStrLn s) >>  runTeletypeIOC   k
+  eff = TeletypeIOC . handleSum (eff . handleCoercible) (\ t -> case t of
+    Read    k -> liftIO getLine      >>= runTeletypeIOC . k
+    Write s k -> liftIO (putStrLn s) >>  runTeletypeIOC   k)
 ```
 
 Here, `ret` is responsible for wrapping pure values in the carrier, and `eff` is responsible for handling an effectful computations. Since the `Carrier` instance handles a sum (`:+:`) of `Teletype` and the remaining signature, `eff` has two parts: a handler for `Teletype` (`alg`), and a handler for teletype effects that might be embedded in other effects in the signature.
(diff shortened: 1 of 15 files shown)


robrix:
I think I’m gonna 🔥 `\/` altogether. There’s no reason to have both it and `handleSum`, and the latter is easier to hold..

src/Control/Effect/Reader.hs
```diff
@@ -56,9 +56,9 @@ newtype ReaderC r m a = ReaderC { runReaderC :: r -> m a }
 
 instance (Carrier sig m, Monad m) => Carrier (Reader r :+: sig) (ReaderC r m) where
   ret a = ReaderC (const (ret a))
-  eff op = ReaderC (\ r -> (alg r \/ eff . handleReader r runReaderC) op)
-    where alg r (Ask       k) = runReaderC (k r) r
-          alg r (Local f m k) = runReaderC m (f r) >>= flip runReaderC r . k
+  eff op = ReaderC (\ r -> handleSum (eff . handleReader r runReaderC) (\case
+    Ask       k -> runReaderC (k r) r
+    Local f m k -> runReaderC m (f r) >>= flip runReaderC r . k) op)
```
robrix: Unfortunately we still need the parentheses around the `LambdaCase`, because of the `op` parameter following it.

src/Control/Effect/Trace.hs
```diff
@@ -55,8 +59,7 @@ newtype TraceByIgnoringC m a = TraceByIgnoringC { runTraceByIgnoringC :: m a }
 
 instance Carrier sig m => Carrier (Trace :+: sig) (TraceByIgnoringC m) where
   ret = TraceByIgnoringC . ret
-  eff = alg \/ (TraceByIgnoringC . eff . handlePure runTraceByIgnoringC)
-    where alg (Trace _ k) = k
+  eff = handleSum (TraceByIgnoringC . eff . handlePure runTraceByIgnoringC) traceCont
```
robrix: This turned out really cute.

README.md
```diff
@@ -195,9 +195,9 @@ newtype TeletypeIOC m a = TeletypeIOC { runTeletypeIOC :: m a }
 instance (Carrier sig m, MonadIO m) => Carrier (Teletype :+: sig) (TeletypeIOC m) where
   ret = TeletypeIOC . ret
 
-  eff = TeletypeIOC . (alg \/ eff . handleCoercible)
-    where alg (Read    k) = liftIO getLine      >>= runTeletypeIOC . k
-          alg (Write s k) = liftIO (putStrLn s) >>  runTeletypeIOC   k
+  eff = TeletypeIOC . handleSum (eff . handleCoercible) (\ t -> case t of
```
robrix: I opted not to use `LambdaCase` in the README to avoid confusing readers unfamiliar with the extension.

PR #48, opened 2018-10-28
Rename to fused-effects

robrix:
This PR renames the package to `fused-effects`, as discussed in #11.


robrix:
I’ve already renamed the repo, so merging.

PR #52, opened 2018-10-29
Once

### LINKED ISSUE #50: AltC Maybe doesn’t terminate for infinitely productive search spaces
```haskell
run (runNonDet (asum (map pure [1..]))) :: Maybe Int
```

won’t terminate despite the fact that it should be able to complete immediately on the first result.

This is probably actually “works as expected,” because of the dual semantics of `Alternative`—there’s no way for `AltC` to know that `<|>` for `Maybe` means “leftmost wins” whereas for `[]` it means “all win.”

Therefore, this probably warrants the inclusion of a `OnceC` which returns in `Maybe` and knows about the `<|>`-as-choice-not-union semantics.

robrix:
- [x] Fixes #50.
- [x] Update the changelog.
- [x] Bump the version.


src/Control/Effect/NonDet.hs
```diff
@@ -30,7 +32,31 @@ instance (Alternative f, Monad f, Traversable f, Carrier sig m, Effect sig, Appl
     Choose k -> liftA2 (<|>) (runAltC (k True)) (runAltC (k False)))
 
 
+-- | Run a 'NonDet' effect, returning the first successful result in an 'Alternative' functor.
+--
+--   Unlike 'runNonDet', this will terminate immediately upon finding a solution.
+--
+--   prop> run (runNonDetOnce (asum (map pure (repeat a)))) == [a]
+--   prop> run (runNonDetOnce (asum (map pure (repeat a)))) == Just a
```
robrix: Note that this is testing it against an infinitely productive search space.

PR #55, opened 2018-10-30
Use PolyKinds in Teletype example, add note to common mistakes

kwf:
Using PolyKinds obviates the need for explicit kind annotations on
type parameters of kinds other than `*` (e.g. Teletype's `m` monad
paramter). This seems to be the more general and less verbose
solution to the kind error issue noted in the docs; I have added a
comment to this effect.

Another thought: perhaps it would be good to have a list in the docs of
language extensions you'll probably need to enable to use this library. In
general, I think this is a useful kind of documentation, which I haven't seen
in many libraries. While Hackage lists the language extensions necessary to
*implement* a library, it doesn't have a mechanism to directly document
which ones you need as an end-user of the library (and for what purposes).

### DIFF
--- a/docs/common_errors.md
+++ b/docs/common_errors.md
@@ -46,3 +46,8 @@ data Teletype (m :: * -> *) k
   deriving (Functor)
 
 ```
+
+Alternatively, to disable this kind-defaulting behavior entirely, use the
+`PolyKinds` language extension. With this extension, the un-annotated version of
+the `Teletype` data type will be inferred to have kind `k -> * -> *`. When you use
+this type, `m`'s kind will correctly be instantiated to `* -> *`.
--- a/examples/Teletype.hs
+++ b/examples/Teletype.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE DeriveFunctor, FlexibleContexts, FlexibleInstances, GeneralizedNewtypeDeriving, KindSignatures, LambdaCase, MultiParamTypeClasses, TypeOperators, UndecidableInstances #-}
+{-# LANGUAGE PolyKinds, DeriveFunctor, FlexibleContexts, FlexibleInstances, GeneralizedNewtypeDeriving, KindSignatures, LambdaCase, MultiParamTypeClasses, TypeOperators, UndecidableInstances #-}
 
 module Teletype where
 
@@ -23,7 +23,7 @@ spec = describe "teletype" $ do
   prop "writes multiple things" $
     \ input output1 output2 -> run (runTeletypeRet input (write output1 >> write output2)) `shouldBe` ((input, [output1, output2]), ())
 
-data Teletype (m :: * -> *) k
+data Teletype m k
   = Read (String -> k)
   | Write String k
   deriving (Functor)


robrix:
One problem with `PolyKinds` is that if you have `PolyKinds` enabled in one file where the effect is defined, and then you define a `Carrier` instance for it in another file, you have to have `PolyKinds` enabled in the second file as well or else you get incredibly-non-obvious kind errors 😞

In that sense, `KindSignatures` are a lot more foolproof. (cf #31)

robrix:
> Another thought: perhaps it would be good to have a list in the docs of
> language extensions you'll probably need to enable to use this library. In
> general, I think this is a useful kind of documentation, which I haven't seen
> in many libraries. While Hackage lists the language extensions necessary to
> *implement* a library, it doesn't have a mechanism to directly document
> which ones you need as an end-user of the library (and for what purposes).

I think this is a superb idea 👍

kwf:
> One problem with PolyKinds is that if you have PolyKinds enabled in one file where the effect is defined, and then you define a Carrier instance for it in another file, you have to have PolyKinds enabled in the second file as well or else you get incredibly-non-obvious kind errors 😞

Oy, I didn't think about that aspect. And it is true that we always want `m` to have the *specific* kind `* -> *`, isn't it?

Might we add a custom type error as a superclass constraint to `Carrier` to diagnose this issue for the user? This may not be possible, and we'd have to be very sure it didn't harm inference, but I'll give it a try when I get a chance.

robrix:
> Oy, I didn't think about that aspect. And it is true that we always want `m` to have the *specific* kind `* -> *`, isn't it?

I believe so. It’s conceivable that there could be cases where `k -> *` would be reasonable, but they’d have to be a little strange: they could have continuations _or_ higher-order positions but not both in the same constructor. And since they always end up being instantiated to `* -> *` in the library, and aren’t really useful outside of effect handlers, I don’t think we gain much from generalizing (aside from inference).

> Might we add a custom type error as a superclass constraint to `Carrier` to diagnose this issue for the user? This may not be possible, and we'd have to be very sure it didn't harm inference, but I'll give it a try when I get a chance.

That’s a wonderful idea—I’d love to see what you come up with!

robrix:
> and aren’t really useful outside of effect handlers

That is, I don’t think it’s likely that effects defined for `fused-effects` will be ever useful in a context where the more general kind would be needed.

kwf:
> Might we add a custom type error as a superclass constraint to Carrier to diagnose this issue for the user? This may not be possible...

I don't believe this is possible. Consider:

```haskell
import GHC.TypeLits hiding (type (*))
import Data.Kind

type family HasKind (j :: Type) (h :: k) :: Constraint where
  HasKind k (h :: k) = ()
  HasKind j (h :: k) =
    TypeError (Text "The type ‘" :<>:
               ShowType h :<>:
               Text "’ should have kind ‘" :<>:
               ShowType j :<>:
               Text "’, but instead it has kind ‘" :<>:
               ShowType k :<>:
               Text "’. " :$$:
               Text "Consider adding an explicit kind annotation.")

-- Note extra constraint in head (everything else unchanged)
--    VVVVVVV
class HasKind ((Type -> Type) -> Type -> Type) h => HFunctor h where
  fmap' :: (a -> b) -> (h m a -> h m b)
  default fmap' :: Functor (h m) => (a -> b) -> (h m a -> h m b)
  fmap' = fmap

  hmap :: (forall x . m x -> n x) -> (h m a -> h n a)
```

Because `HFunctor` has methods whose type signatures constrain the kind of `h`, we can't generalize the kind of `h` by saying `class ... => HFunctor (h :: k) where ...`, as this doesn't kind-check with the more-specific-kinded body. The constraint solver only examines the instance head once it's found a matching instance, which means that without this generalization, a matching instance can't be selected because the mere idea of one is a kind error—so we never see the custom type error. This might work for a class whose methods were kind-polymorphic re:`h`, but in this case I don't see how to get the custom type error to fire before the kind error itself. :(

kwf:
> Because `HFunctor` has methods whose type signatures constrain the kind of `h`, we can't generalize the kind of `h` by saying `class ... => HFunctor (h :: k) where ...`, as this doesn't kind-check with the more-specific-kinded body.

What I implicitly mean by generalizing the kind of `h` by annotating it like `(h :: k)` is that PolyKinds would be enabled here, making `k` a polymorphic kind variable.

kwf:
Indeed, I thought the below would work, but even it doesn't:

```haskell
class HasKind ((Type -> Type) -> Type -> Type) h => CouldBeHFunctor h

class CouldBeHFunctor h => HFunctor h where
  ...
```

This implies that the user has to write the boilerplate `instance CouldBeHFunctor MyType`, but my hope was that this hand-written instance would trigger the custom error at its instance declaration site, and this would be reported before (or with) the underlying kind error—no such luck. I think this is due to GHC's multi-phase approach to constraint solving for typeclasses. You have to solve all the kind errors before you can start seeing instance head constraint errors.

robrix:
> This implies that the user has to write the boilerplate `instance CouldBeHFunctor MyType`, but my hope was that this hand-written instance would trigger the custom error at its instance declaration site, and this would be reported before (or with) the underlying kind error—no such luck. I think this is due to GHC's multi-phase approach to constraint solving for typeclasses. You have to solve all the kind errors before you can start seeing instance head constraint errors.

That’s a shame, but a heroic effort nonetheless!

robrix:
`KindSignatures` or `PolyKinds` are both much less necessary as of #184. Between that and the challenges noted above, @kwf and I have agreed to close this out.

Thanks very much @kwf for the exploration and the thoughtful discussion! 💖

PR #58, opened 2018-11-01
[WIP] Cut effect

robrix:
This PR defines a `Cut` effect, suitable for adding committed choice to nondeterminism effects.

- [x] Defines a `Cut` effect.
- [x] Defines a handler for `Cut` which prunes nondeterminism following cut branches.
- [ ] 📝 & test the new stuff.
- [ ] Bump the changelog.
- [ ] Bump the package version.

### DIFF
--- a/src/Control/Effect/Cut.hs
+++ b/src/Control/Effect/Cut.hs
@@ -1,0 +1,1 @@
+module Control.Effect.Cut where
--- a/src/Control/Effect/Cut.hs
+++ b/src/Control/Effect/Cut.hs
@@ -4,6 +4,7 @@ module Control.Effect.Cut
 , cutfail
 , call
 , cut
+, Branch(..)
 ) where
 
 import Control.Applicative (Alternative(..))
@@ -32,3 +33,10 @@ call m = send (Call m ret)
 
 cut :: (Alternative m, Carrier sig m, Member Cut sig) => m ()
 cut = pure () <|> cutfail
+
+
+data Branch a
+  = Prune
+  | None
+  | Some a
+  deriving (Eq, Functor, Ord, Show)
--- a/src/Control/Effect/Cut.hs
+++ b/src/Control/Effect/Cut.hs
@@ -1,9 +1,12 @@
-{-# LANGUAGE DeriveFunctor, ExistentialQuantification, StandaloneDeriving #-}
+{-# LANGUAGE DeriveFunctor, ExistentialQuantification, FlexibleContexts, StandaloneDeriving #-}
 module Control.Effect.Cut
 ( Cut(..)
+, cutfail
+, call
 ) where
 
 import Control.Effect.Carrier
+import Control.Effect.Sum
 
 data Cut m k
   = Cutfail
@@ -18,3 +21,9 @@ instance HFunctor Cut where
 instance Effect Cut where
   handle _     _       Cutfail    = Cutfail
   handle state handler (Call m k) = Call (handler (m <$ state)) (handler . fmap k)
+
+cutfail :: (Carrier sig m, Member Cut sig) => m a
+cutfail = send Cutfail
+
+call :: (Carrier sig m, Member Cut sig) => m a -> m a
+call m = send (Call m ret)
--- a/src/Control/Effect/Cut.hs
+++ b/src/Control/Effect/Cut.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE DeriveFunctor, ExistentialQuantification, FlexibleContexts, StandaloneDeriving #-}
+{-# LANGUAGE DeriveFunctor, ExistentialQuantification, FlexibleContexts, FlexibleInstances, LambdaCase, MultiParamTypeClasses, StandaloneDeriving, TypeOperators, UndecidableInstances #-}
 module Control.Effect.Cut
 ( Cut(..)
 , cutfail
@@ -6,10 +6,14 @@ module Control.Effect.Cut
 , cut
 , Branch(..)
 , branch
+, runCut
+, CutC(..)
 ) where
 
 import Control.Applicative (Alternative(..))
 import Control.Effect.Carrier
+import Control.Effect.Internal
+import Control.Effect.NonDet
 import Control.Effect.Sum
 
 data Cut m k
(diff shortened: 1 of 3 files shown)


src/Control/Effect/Cut.hs
```diff
@@ -0,0 +1,69 @@
...
+, CutC(..)
+) where
+
+import Control.Applicative (Alternative(..))
+import Control.Effect.Carrier
+import Control.Effect.Internal
+import Control.Effect.NonDet
+import Control.Effect.Sum
+
+data Cut m k
```
aymannadeem: Just to ensure I'm understanding correctly, the `Cut` effect is described in Effect Handlers in Scope (even though our handler doesn't look anything like the one in literature). `Cut` has thse two operations: 
1. `Cutfail`, a hard fail
2. `Call`, which delimits the scope _where_ `Cutfail` should fail within

`Cut` is committed choice, and the entire purpose is so that once we've matched the branch we want, we don't try any other branch. This is a useful capability since we don't have to backtrack (backtracking is quadratic). 

You want to do different thing when you `Fail` vs. `Cutfail` because `Cutfail` is when you stop pursuing other branches. 

@robrix is this why we don't do the table lookup in our other branch? Because we're getting rid of the need to do so (which requires maintaining a representation of trees) or am I getting things mixed up?

robrix: You’ve got it 👍

> @robrix is this why we don't do the table lookup in our other branch? Because we're getting rid of the need to do so (which requires maintaining a representation of trees) or am I getting things mixed up?

The way we were doing the table-based choice thing depended on having a representation of the computation as a tree that it could inspect, yes. It’s not clear how to do that with `fused-effects`, but it’s likely possible to recover that behaviour with some creativity. But I’m hoping that the efficiency gains from using `fused-effects` in the first place will outweigh the cost of not doing the efficient table-based choice, and therefore that we won’t have a pressing need to reimplement tabling.

robrix:
I had intended to bump the package version in this PR, but upon reflection I’ve decided to tackle a couple of other things before the next release. Therefore, I’ve added a note to the changelog, but haven’t changed the version.

src/Control/Effect/Cut.hs
```diff
@@ -0,0 +1,100 @@
...
+) where
+
+import Control.Applicative (Alternative(..))
+import Control.Effect.Carrier
+import Control.Effect.Internal
+import Control.Effect.NonDet
+import Control.Effect.Sum
+
+-- | 'Cut' effects are used with 'NonDet' to provide control over backtracking.
+data Cut m k
```
robrix: As discussed with @aymannadeem, a `Cut` effect is described in _Effect Handlers in Scope_ §5, as an example of a non-orthogonal effect: i.e. an effect that interacts with other effects (in this case, nondeterminism).

src/Control/Effect/Cut.hs
```diff
@@ -0,0 +1,100 @@
...
+instance Effect Cut where
+  handle _     _       Cutfail    = Cutfail
+  handle state handler (Call m k) = Call (handler (m <$ state)) (handler . fmap k)
+
+-- | Fail the current branch, and prevent backtracking within the nearest enclosing 'call' (if any).
+--
+--   Contrast with 'empty', which fails the current branch but allows backtracking.
+--
+--   prop> run (runNonDetOnce (runCut (cutfail <|> pure a))) == Nothing
+--   prop> run (runNonDetOnce (runCut (pure a <|> cutfail))) == Just a
```
robrix: I think these tests sum up `cutfail` very nicely.

While I documented the contrast with `empty`, I opted not to test it, as it’s adequately covered in the tests of `runNonDet`.

src/Control/Effect/Cut.hs
```diff
@@ -0,0 +1,100 @@
...
+cutfail :: (Carrier sig m, Member Cut sig) => m a
+cutfail = send Cutfail
+
+-- | Delimit the effect of 'cutfail's, allowing backtracking to resume.
+--
+--   prop> run (runNonDetOnce (runCut (call (cutfail <|> pure a) <|> pure b))) == Just b
```
robrix: Likewise, this test for `call`’s interaction with `cutfail` is pleasantly simple.

src/Control/Effect/Cut.hs
```diff
@@ -0,0 +1,100 @@
...
+
+instance (Alternative m, Carrier sig m, Effect sig, Monad m) => Carrier (Cut :+: NonDet :+: sig) (CutC m) where
+  ret = CutC . ret . Some
+  eff = CutC . handleSum (handleSum
+    (eff . handle (Some ()) (branch (ret Prune) (ret None) runCutC))
+    (\case
+      Empty    -> ret None
+      Choose k -> runCutC (k True) >>= branch (ret Prune) (runCutC (k False)) (\ a -> ret (Some a) <|> runCutC (k False))))
+    (\case
+      Cutfail  -> ret Prune
```
robrix: I was initially unsure how to implement a carrier for `Cut` given that the paper’s handler uses a function which can’t be fit to the shape of the distributive law necessary for handling `Cut` effects embedded in other effects (specifically, it can’t be made rank-_n_). But the essential characteristic is that there’s a way to distinguish `Cutfail` from `Empty`, and that handling of `Choose` respects the difference.

By representing the carrier’s results with a `Maybe`-like functor with two cases for nothing, we have exactly the necessary properties. Note that we could also have used `Either Bool a`, and `either (bool a b) c` in place of `branch a b c`. On the balance, I feel that clear, purposeful names are valuable enough to warrant the additional surface area.

src/Control/Effect/Cut.hs
```diff
@@ -0,0 +1,100 @@
...
+      Call m k -> runCutC m >>= branch (ret None) (ret None) (runCutC . k))
+
+
+-- $setup
+-- >>> :seti -XFlexibleContexts
+-- >>> import Test.QuickCheck
+-- >>> import Control.Effect.Void
+-- >>> instance Arbitrary a => Arbitrary (Branch a) where arbitrary = frequency [(1, pure Prune), (1, pure None), (3, Some <$> arbitrary)] ; shrink b = case b of { Some a -> Prune : None : map Some (shrink a) ; None -> [Prune] ; Prune -> [] }
```
robrix: I selected these frequencies based on the frequencies selected in the `Arbitrary1` instance for `Maybe`: `1` for `Nothing`, `3` for `Just`.

PR #60, opened 2018-11-03
Reformat the license

robrix:
This PR reformats the `LICENSE` file in an attempt to get a pretty license banner like https://github.com/atom/atom/blob/master/LICENSE.md does.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,30 +1,27 @@
 Copyright (c) 2018, Rob Rix and Patrick Thomson
-
 All rights reserved.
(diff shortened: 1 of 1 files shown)


robrix:
I used the output of the license template this time, so it _should_ work. I think maybe it only shows up on `master`, so I’m going to merge this. We can tweak further as necessary.

robrix:
[It works!](https://github.com/robrix/fused-effects/blob/master/LICENSE)

PR #63, opened 2018-11-03
Add Nicolas Wu & Tom Schrijvers to authors & copyrights

robrix:
Nicolas Wu’s & Tom Schrijvers’ work was monumentally important for this development; aside from a few things being renamed, a slight tweak to the shape of effect functors to integrate _Effect Handlers in Scope_ with _Fusion for Free_, and one or two new effects, this work was derived directly from their papers. Therefore, I’m adding them to the LICENSE and `authors` field.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 BSD 3-Clause License
 
-Copyright (c) 2018, Rob Rix and Patrick Thomson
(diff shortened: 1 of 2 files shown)


robrix:
Got verbal approval from @patrickt too.

PR #66, opened 2018-11-04
Cut complexity

### LINKED ISSUE #65: runCut creates exponential right-nested branches
`runCut` interprets `NonDet` into an underlying `Alternative` in a way that ends up creating exponential numbers of right-nested branches. Therefore, the problem doesn’t show up with `runNonDetOnce`, as it terminates immediately on finding a solution in a left-to-right search, but with `runNonDet` (which explores the entire search space) it’s quite apparent.

robrix:
This PR:

- [x] Adds an `Alt` constructor to `Branch`, and constructs `Alt`s as the interpretation of `Choose`.
- [x] Allows `Cut` to be handled independent of the search strategy for the underlying `Alternative` instance. E.g. `runCut` now composes well with both `runNonDet` and `runNonDetOnce`.
- [x] Fixes #65.

### DIFF
--- a/src/Control/Effect/Cut.hs
+++ b/src/Control/Effect/Cut.hs
@@ -81,9 +81,6 @@ branch _ _ _ f (Alt a b) = f a b
 runBranch :: Alternative m => Branch m a -> m a
 runBranch = branch empty empty pure (<|>)
 
-bindBranch :: (Alternative m, Carrier sig m, Monad m) => m (Branch m a) -> (b -> m (Branch m a)) -> Branch m b -> m (Branch m a)
-bindBranch cut bind = branch cut (ret None) bind (\ a b -> ret (Alt (a >>= bind >>= runBranch) (b >>= bind >>= runBranch)))
-
 
 -- | Run a 'Cut' effect within an underlying 'Alternative' instance (typically 'Eff' carrying a 'NonDet' effect).
 --
@@ -103,6 +100,8 @@ instance (Alternative m, Carrier sig m, Effect sig, Monad m) => Carrier (Cut :+:
     (\case
       Cutfail  -> ret Cut
       Call m k -> runCutC m >>= bindBranch (ret None) (runCutC . k))
+    where bindBranch :: (Alternative m, Carrier sig m, Monad m) => m (Branch m a) -> (b -> m (Branch m a)) -> Branch m b -> m (Branch m a)
+          bindBranch cut bind = branch cut (ret None) bind (\ a b -> ret (Alt (a >>= bind >>= runBranch) (b >>= bind >>= runBranch)))
 
 
 -- $setup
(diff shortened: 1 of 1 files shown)


src/Control/Effect/Cut.hs
```diff
@@ -48,53 +49,64 @@ call m = send (Call m ret)
...
 instance (Alternative m, Carrier sig m, Effect sig, Monad m) => Carrier (Cut :+: NonDet :+: sig) (CutC m) where
-  ret = CutC . ret . Some
+  ret = CutC . ret . Pure
   eff = CutC . handleSum (handleSum
-    (eff . handle (Some ()) (branch (ret Prune) (ret None) runCutC))
+    (eff . handle (Pure ()) (bindBranch (ret Cut) runCutC))
     (\case
       Empty    -> ret None
-      Choose k -> runCutC (k True) >>= branch (ret Prune) (runCutC (k False)) (\ a -> ret (Some a) <|> runCutC (k False))))
+      Choose k -> runCutC (k True) >>= branch (ret Cut) (runCutC (k False)) (\ a -> ret (Alt (ret a) (runCutC (k False) >>= runBranch))) (fmap ret . Alt)))
```
robrix: We now handle `Choose` by constructing `Alt`s, deferring choices to the underlying `Alternative` instance without building exponentially many branches.

src/Control/Effect/Cut.hs
```diff
@@ -48,53 +49,64 @@ call m = send (Call m ret)
...
 
 -- | The result of a nondeterministic branch of a computation.
-data Branch a
-  = Prune
+data Branch m a
+  = Cut
   | None
-  | Some a
-  deriving (Eq, Functor, Ord, Show)
+  | Pure a
```
robrix: I renamed a couple of the `Branch` constructors to better describe the operations they represent.

src/Control/Effect/Cut.hs
```diff
@@ -48,53 +49,64 @@ call m = send (Call m ret)
 
 -- | Commit to the current branch, preventing backtracking within the nearest enclosing 'call' (if any) on failure.
 --
---   prop> run (runNonDet (runCut (cut *> pure a <|> pure b))) == [a, b]
+--   prop> run (runNonDet (runCut (pure a <|> cut *> pure b))) == [a, b]
+--   prop> run (runNonDet (runCut (cut *> pure a <|> pure b))) == [a]
```
robrix: Importantly, `cut` now prunes when successful as well. This is the behaviour of `cut` in Hinze’s _Deriving Backtracking Monad Transformers_, and I believe it’s correct w.r.t. our desired semantics: e.g. in a parser, you want to avoid backtracking after a successful `cut` as well as a failed one in order to limit ambiguity.

Note that in Kiselyov et al’s _Backtracking, Interleaving, and Terminating Monad Transformers_, they discuss Prolog-style _cut_’s conflation of negation with pruning; this criticism applies to `Cut` as well. However, it’s reasonable to consider adding these as orthogonal features using a new effect, and in the meantime, `cut` is a reasonably well-understood logical operator.

PR #67, opened 2018-11-04
Parsing example

robrix:
This PR develops an example of using `fused-effects` to write parsers, following _Effect Handlers in Scope_, §6. In particular, it shows the combination of `NonDet`, `Cut`, and a custom `Symbol` effect to implement a grammar for simple arithmetic expressions (lacking whitespace).

### DIFF
--- a/examples/Parser.hs
+++ b/examples/Parser.hs
@@ -1,0 +1,1 @@
+module Parser where
--- a/examples/Parser.hs
+++ b/examples/Parser.hs
@@ -13,6 +13,10 @@ import Test.QuickCheck
 
 spec :: Spec
 spec = describe "parser" $ do
+  describe "parse" $ do
+    prop "returns pure values at the end of input" $
+      \ a -> run (runNonDetOnce (parse "" (pure a))) == Just (a :: Integer)
+
   describe "satisfy" $ do
     prop "matches with a predicate" $
       \ c f -> run (runNonDetOnce (parse [c] (satisfy (applyFun f)))) == if applyFun f c then Just c else Nothing
--- a/examples/Parser.hs
+++ b/examples/Parser.hs
@@ -1,5 +1,12 @@
 {-# LANGUAGE DeriveFunctor, KindSignatures #-}
 module Parser where
 
+import Control.Effect.Carrier
+import Data.Coerce
+
 data Symbol (m :: * -> *) k = Symbol Char (Char -> k)
   deriving (Functor)
+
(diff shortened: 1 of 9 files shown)


examples/Parser.hs
```diff
@@ -0,0 +1,134 @@
...
+      \ a -> run (runNonDet (runCut (parse (show (abs a)) expr))) == [abs a]
+
+    prop "matches multiplication" $
+      \ as -> run (runNonDet (runCut (parse (intercalate "*" (show . abs <$> 1:as)) expr))) == [product (map abs as)]
+
+    prop "matches addition" $
+      \ as -> run (runNonDet (runCut (parse (intercalate "+" (show . abs <$> 0:as)) expr))) == [sum (map abs as)]
+
+    prop "respects order of operations" . forAll (sized arbNested) $
+      \ as -> run (runNonDet (runCut (parse (intercalate "+" (intercalate "*" . map (show . abs) . (1:) <$> [0]:as)) expr))) == [sum (map (product . map abs) as)]
```
robrix: This doesn’t produce arbitrarily-nested expressions, merely sums of products, but it’s still an interesting test.

examples/Parser.hs
```diff
@@ -0,0 +1,134 @@
...
+      c:cs | p c -> runParseC (k c) cs
+      _          -> empty)
+    op)
+  {-# INLINE eff #-}
+
+
+expr :: (Alternative m, Carrier sig m, Member Cut sig, Member Symbol sig, Monad m) => m Int
+expr = do
+  i <- term
+  call ((i +) <$ char '+' <* cut <*> expr
```
robrix: We use `cut` and `call` to commit to the branch after matching the operator.

examples/Parser.hs
```diff
@@ -0,0 +1,134 @@
...
+
+    prop "respects order of operations" . forAll (sized arbNested) $
+      \ as -> run (runNonDet (runCut (parse (intercalate "+" (intercalate "*" . map (show . abs) . (1:) <$> [0]:as)) expr))) == [sum (map (product . map abs) as)]
+
+    where arbNested :: Arbitrary a => Int -> Gen [[a]]
+          arbNested 0 = pure []
+          arbNested n = do
+            Positive m <- arbitrary
+            let n' = n `div` (m + 1)
+            replicateM m (vector n')
```
robrix: Without this, we were generating _extremely_ large nested lists.

PR #78, opened 2018-11-28
Fix the ghc-8.2 doctests

robrix:
This PR aims to resolve the doctest build errors we’re seeing for ghc 8.2.


robrix:
Aha! Looks like always writing ghc environment files was the missing ingredient.

That said, I think it’s super reckless of this script to be 🔥ing the `cabal.project` file, so I’m going to try using [Travis’ builtin Haskell support](https://github.com/fedora-haskell/fedora-haskell-tools/blob/master/.travis.yml) instead.

robrix:
Alright, this is in pretty good shape. Ready for review!

PR #81, opened 2018-12-03
Loosen QuickCheck bounds

robrix:
Per https://github.com/commercialhaskell/stackage/pull/4139.


robrix:
https://hackage.haskell.org/package/fused-effects-0.1.2.1 published.

PR #91, opened 2019-01-05
Writer functionality

### LINKED ISSUE #83: Writer effects lack useful operations
`WriterT` & `MonadWriter` offer a bunch of useful operations besides `tell` (types adapted to fit `fused-effects` conventions):

```haskell
writer :: (Carrier sig m, Member (Writer w) sig) => (a, w) -> m a
listen :: (Carrier sig m, Member (Writer w) sig) => m a -> m (w, a)
listens :: (Carrier sig m, Member (Writer w) sig) => (w -> b) -> m a -> m (b, a)
pass :: (Carrier sig m, Member (Writer w) sig) => m (w -> w, a) -> m a
censor :: (Carrier sig m, Member (Writer w) sig) => (w -> w) -> m a -> m a
```

`mtl` requires `MonadWriter` instances to provide either `tell` or `writer`, as well as `listen` and `pass`, defining `listens` in terms of `listen` and `censor` in terms of `pass`. Based on that, I’m reasonably sure that `pass` would need to be a scoped operation, and probably `listen` would as well, since it ranges over the output of its argument.

robrix:
This PR:

- [x] Depends on #90.
- [x] Implements `censor`, allowing local modification of a computation’s output.
- [x] Implements `listen`, allowing observation of a computation’s output.
- [x] Implements `listens`, allowing mapped observation of a computation’s output.
- [x] Fixes #83.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,3 +1,4 @@
+- Adds `listen`, `listens`, and `censor` operations to `Writer`.
 - Adds benchmarks of `WriterC`/`VoidC` wrapped with `Eff` against their unwrapped counterparts.
 - Adds `Functor`, `Applicative`, and `Monad` instances for `WriterC`.
 - Adds `Functor`, `Applicative`, and `Monad` instances for `VoidC`.
--- a/src/Control/Effect/Writer.hs
+++ b/src/Control/Effect/Writer.hs
@@ -49,7 +49,7 @@ listen m = send (Listen m (curry ret))
 
 -- | Run a computation, applying a function to its output and returning the pair of the modified output and its result.
 --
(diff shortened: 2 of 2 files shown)


robrix:
In `mtl`, `censor` is defined in terms of `pass :: m (a, w -> w) -> m a`. Implementing that operation would require some pretty significant changes to `handle`, strengthening the `Functor` constraint on `f` to `Traversable`, and introducing a `Functor` constraint on `n`. That in turn would require a bunch of `Carrier` instances to add extra constraints. I’m not going to outright reject that possibility, but I think I’d rather wait til we or someone else needs `pass` rather than do it preemptively.

Issue #96, opened 2019-01-13
Propagating MonadIO constraint downwards


deepfire:
How does one use the underlying monad that is `MonadIO`-constrained using `runM`?

On the first approach it doesn't work (full exhibit at [1]):

```
doliftio :: MonadIO m => m ()
doliftio = runM (liftIO $ putStr "")
```
..yields..
```
test/Control/Effect/LiftIO.hs:26:18: error:
    â¢ Could not deduce (Member (Lift IO) (Lift m))
        arising from a use of âliftIOâ
      from the context: MonadIO m
        bound by the type signature for:
                   doliftio :: forall (m :: * -> *). MonadIO m => m ()
        at test/Control/Effect/LiftIO.hs:25:1-29
    â¢ In the first argument of ârunMâ, namely â(liftIO $ putStr "")â
      In the expression: runM (liftIO $ putStr "")
      In an equation for âdoliftioâ: doliftio = runM (liftIO $ putStr "")
   |
28 | doliftio = runM (liftIO $ putStr "")
   |                  ^^^^^^^^^^^^^^^^^^
```

--
1. https://github.com/deepfire/fused-effects/blob/2a10f9351d8602b6c32b7af8830d58c76dad0655/test/Control/Effect/LiftIO.hs

robrix:
`Lift IO` gives `Eff` a `MonadIO` instance, but `runM` takes `Eff (LiftC m) a -> m a`, and not e.g. `MonadIO m => Eff (LiftC IO) a -> m a`. Therefore, since `runM` should return in `IO`, you’ll need to add another `liftIO` at the outside:

```haskell
doliftio :: MonadIO m => m ()
doliftio = liftIO (runM (liftIO (putStrLn "hello")))
```

or equivalently, change the type of `doliftio` to `IO`:

```haskell
doliftio :: IO ()
doliftio = runM (liftIO (putStrLn "hello"))
```

robrix:
A closely related question is why we only provide a `MonadIO` instance for `Lift IO` and not for e.g. `MonadIO m => Lift m`, and it’s basically down to ambiguity; the ergonomics are much worse if using `liftIO` in an `Eff` context doesn’t fix the type of `m` to `IO`, because the type checker would essentially have to guess `m` at every point where you’re using `liftIO`.

[...]

PR #97, opened 2019-01-13
[RFC] An attempt to provide transparent way to use `liftIO`

deepfire:
This is an attempt to solve #96.

There is probably much not to like about it -- notably, the `runM` no longer provides a `MonadIO` instance -- it's instead subsumed by `runMIO`.

I couldn't find any other way to solve the problem elucidated by the test.

Any better option is very welcome!


robrix:
@deepfire: Thanks for the PR! This is a creative solution, but I’d prefer to avoid having two different monad-lifting effects.

That said, it does seem like it’d be possible to use `LiftIOC` as a carrier for `Lift IO` with essentially the same definition that it currently has—`LiftIO` isn’t specialized to `IO`, after all. A carrier with those semantics would still lack the other side of this problem (i.e. lifting both `IO` and non-`IO` monadic computations into the same context requires you to use `Lift m` for both and thus lose the `MonadIO` instance), but might be slightly more convenient to use. But since [you can use `liftIO` outside of the call to `runM`](
https://github.com/robrix/fused-effects/issues/96#issuecomment-453837657) to achieve the same thing, I’m not sure that it’s worth adding this to the lib.

Happy to hear your thoughts on the matter, tho!

deepfire:
I'm happy there is a more general solution (I was just fumbling, really..) -- and happy to close the PR : -)

PR #114, opened 2019-02-27
Nondeterminism rework

robrix:
This PR reworks the carriers for nondeterminism and related effects to be more efficient/better-behaved monad transformers.

- [x] Depends on #111.
- [x] Replaces `AltC` with a monad transformer which does not require commutativity for correctness.
- [x] Replaces `CutC` with a monad transformer implemented using the recipe laid out by Hinze in _Deriving Backtracking Monad Transformers_.
- [x] Replaces `CullC` similarly.
- [x] 🔥s `Branch`.
- [x] 📝 `NonDetC`, referencing _Deriving Backtracking Monad Transformers_.
- [x] Update the changelog.
- [x] Fixes #116.

PR #119, opened 2019-03-01
Generic deriving for HFunctor and Effect

isovector:
I used @kcsongor's [approach for deriving bifunctor](https://kcsongor.github.io/generic-deriving-bifunctor/) to give generic implementations for `HFunctor` and `Effect`. 

Unfortunately we can't get `Generic` instances for effects that have existential parameters :(

~TBH I am not super sure that these `INCOHERENT` instances are safe. But if they are, and~ if you like this idea, I'll cleanup the code and write some tests to ensure everything inlines away.

### DIFF
--- a/examples/Teletype.hs
+++ b/examples/Teletype.hs
@@ -1,9 +1,13 @@
+{-# LANGUAGE DeriveAnyClass                                                                                                                                                         #-}
 {-# LANGUAGE DeriveFunctor, FlexibleContexts, FlexibleInstances, GeneralizedNewtypeDeriving, KindSignatures, LambdaCase, MultiParamTypeClasses, TypeOperators, UndecidableInstances #-}
+{-# LANGUAGE DeriveGeneric                                                                                                                                                          #-}
+{-# LANGUAGE DerivingStrategies                                                                                                                                                     #-}
 
 module Teletype where
 
 import Prelude hiding (read)
 
+import GHC.Generics (Generic)
 import Control.Effect
 import Control.Effect.Carrier
 import Control.Effect.Sum
@@ -26,15 +30,10 @@ spec = describe "teletype" $ do
 data Teletype (m :: * -> *) k
   = Read (String -> k)
   | Write String k
-  deriving (Functor)
+  deriving (Functor, Generic)
 
-instance HFunctor Teletype where
-  hmap _ = coerce
-  {-# INLINE hmap #-}
-
-instance Effect Teletype where
-  handle state handler (Read    k) = Read (handler . (<$ state) . k)
-  handle state handler (Write s k) = Write s (handler (k <$ state))
+instance HFunctor Teletype
+instance Effect Teletype
 
 read :: (Member Teletype sig, Carrier sig m) => m String
 read = send (Read ret)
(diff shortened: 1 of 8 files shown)


kcsongor:
I think the `INCOHERENT` instances are safe, here's why:
when GHC tries to derive a function arising from a default signature, it will instantiate all type variables with *distinct* skolems. In the case of `hmap`, `m`, `n` and `a` are thus certainly different types, say `m0`, `n0` and `a0`. We can think of these skolems as *concrete* types.

This means that the instance can be read as
`instance {-# INCOHERENT #-} GHFunctor (Rec0 (m0 x)) (Rec0 (n0 x)) m0 n0 where` (1)
instance really only matches the right `m0` and `n0` types.

With this in mind, the fields that don't mention `m` will only match the last instance
`instance {-# INCOHERENT #-} GHFunctor (K1 _1 z) (K1 _1 z) m0 n0 where` (2)
so that's fine. These fields will guaranteed to match, because they have the same type in both `Rep (h m a)` and `Rep (h n a)`, as they don't refer to the first parameter.

Similarly, this instance works out just fine:
`instance {-# INCOHERENT #-} HFunctor l => GHFunctor (Rec0 (l m0 x)) (Rec0 (l n0 x)) m0 n0 where` (3)

We might think that (1) and (3) overlap and that might cause trouble, but that's not possible, because that would require that `l m0 ~ m0`, which is can't happen due to the occurs check.

This all works out is because the `hmap` function is derived only once, and not at every call site. For this reason, it is impossible that the distinct type variables will be instantiated to the same type.

The bottom line is that when using this pattern (default signatures + incoherent instances), we are safe to treat the universals as distinct types.

robrix:
Thanks for the PR! I’m a bit swamped with #111 & #114 at the moment, but I’ll try to give this more thorough attention shortly.

isovector:
So I'm deeper in my understanding of `fused-effects` and I'm not convinced this PR is entirely wise. The `HFunctor` generic implementation is fine---naturality means this implementation must be unique. However, instances for `Effect` are less well-founded.

Something that became only obvious to me after reading "Effect Handlers in Scope" is that the majority of the work actually occurs inside of `handle`. Higher-order effects need to think carefully about their `handle` instance, since many implementations are possible The instances given by this PR are correct for first-order handlers, but those are given for free with a `Functor` instance via `Lift` anyway. 

I wonder if this might be more of a documentation problem than anything else; [docs/defining_effects](https://github.com/robrix/fused-effects/blob/master/docs/defining_effects.md) makes no mention of higher-order effects, nor does it mention `Lift`.

robrix:
@isovector: I’m inclined to agree; an interesting example is the `Error` effect’s `Catch` constructor which has two higher-order actions (one directly embedded, the other the result of the handler) and which uses the same suspended state for each (i.e. rolling back changes on throw); contrast with `Resource`, which passes the state constructed by the acquire action on to the release and use actions.

I’m generally inclined not to embed first-order actions using `Lift`; while it’s definitely possible (e.g. we could define `State s` using `Lift (State s)`), it feels unnecessary to make the distinction between first- and higher-order effects at the interface boundaries. But for first-order effects, we could also possibly give default implementations of `hmap` and `handle` using `DefaultSignatures` and a `Coercible` constraint.

PR #120, opened 2019-03-04
Rename Void to Pure

### LINKED ISSUE #117: VoidC is a terrible name
`VoidC` has one and only one job: providing `run`. We should name it `PureC` or `RunC` or something instead of `VoidC`.

robrix:
This PR renames `Control.Effect.Void`, `Void`, and `VoidC` to `Control.Effect.Pure`, `Pure`, and `PureC`, respectively.

- [x] Fixes #117. Note that while `PureC` is indeed the identity functor, we’ve opted for the name `Pure` to try to emphasize that it represents the absence of effects, and that the only thing it can do (modulo ⊥, etc.) is return pure result values.


src/Control/Effect/Pure.hs
```diff
@@ -1,50 +1,50 @@
...
 
 -- | Run an action exhausted of effects to produce its final result value.
-run :: VoidC a -> a
-run = runVoidC
+run :: PureC a -> a
+run = runPureC
 {-# INLINE run #-}
 
-newtype VoidC a = VoidC { runVoidC :: a }
+newtype PureC a = PureC { runPureC :: a }
```
robrix: We can attach docs to the selector, and thus name the selector `run` instead of using a separate top-level declaration for it. On the other hand, this is pleasantly consistent, and I’m not sure if it’s really worth the four or so lines saved.

PR #125, opened 2019-03-07
Experiment with doing builds directly in GitHub Actions

tclem:
This is an experiment to play with doing builds in GitHub Actions. I tried a number of different things here:

1. Create a dedicated [cabal action](https://github.com/robrix/cabal-action) and use that in a workflow to do the builds.
2. Create an action in this repo that setups up the build env and attempts to pre-build dependencies. The workflow initiates the actual project build.
3. Create an action that does the entire build.

It's a little bit funny, but I think option 3 is going to be the best because it allows taking advantage of layer caching in docker. A completely cold build is about 15 mins, but after that a change to just fused-effects only takes 1.5 mins to build. With the other two options I found that things like `cabal new-update` and `cabal new-build` were having to run again at workflow execution time (essentially a `docker run`). In the last option, everything happens at `docker build` time so we get to use layer caches as appropriate. The funny part is that in the actual workflow we don't need the action to do *anything* more so it's kind of a noop at that point.


tclem:
Interesting, but stale! Going to close out.

PR #178, opened 2019-06-13
Uses generic derivation for reinterpret log example

jkachmar:
The `ReinterpretLog` example had a comment mentioning that generic derivation of `HFunctor` and `Effect` for first-order effects wasn't possible at the time, and linked to an issue (#54) that has since been closed (#170).

This PR updates the example with a generic derivation for these effects and removes the comment, which is no longer applicable.

### DIFF
--- a/examples/ReinterpretLog.hs
+++ b/examples/ReinterpretLog.hs
@@ -10,6 +10,7 @@
 --   structured log messages as strings.
 
 
+{-# LANGUAGE DeriveAnyClass             #-}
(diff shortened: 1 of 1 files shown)


jkachmar:
Funny thing is that I was originally going to try and create some `fused-effects-th` package to automatically gin up these instances + the helper functions (like `polysemy` does) when I looked at the `Teletype` example and realized that you'd already built-in a way nicer solution for the instances stuff 😂

PR #184, opened 2019-07-06
Constrained effects

robrix:
This PR changes the signature of `eff` from `sig m (m a) -> m a` to `sig m a -> m a`, bringing it closer into line with the presentations in the literature, and making `eff` much more obviously a `sig`-algebra, and enabling interoperation with e.g. `bound` and higher-order free monads in general.

- [x] Simplifies the types of `eff`, `send` and all the effect constructors to allow for much simpler algebraic usage and enable integration with other presentations of free monads (e.g. syntax à la `bound`).
- [x] Adds constraints to `hmap` and `handle` to enable more defining instances.
- [x] ~~Removes~~ _Deprecates_ `fmap'`. (cf #156.) I got cold feet, so deprecated it instead of deleting it.
- [x] Deprecates `handlePure`.
- [x] Bumps the package version to 0.5 to match the changelog (so I can select this version unambiguously in some local experiments even before pushing to hackage).
- [x] Replaces the default definitions of `hmap` and `handle` with generically-derived ones (since coercion is no longer possible).

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,7 @@
 # v0.5.0.0
 
+- Derives `Generic1` instances for all non-existentially-quantified effect datatypes.
+
 ## Backwards-incompatible changes
 
 - Replaces `runResource` with an equivalent function that uses `MonadUnliftIO` to select the correct unlifting function (a la `withResource`, which is removed in favor of `runResource`).
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -8,6 +8,8 @@
 
 - Changes the signature of `eff` from `sig m (m a) -> m a` to `sig m a -> m a`, requiring effects to hold `m k` in their continuation positions instead of merely `k`. This was done in order to improve interoperability with other presentations of higher-order syntax, e.g. `bound`; syntax used with `bound` can now be given `HFunctor` and `Carrier` instances.
 
+  To upgrade effects used with previous versions, change any continuations from `k` to `m k`. If no existential type variables appear in the effect, you can derive `Generic1`, and thence `HFunctor` & `Effect` instances. Otherwise, implement the required instances by hand. Since continuation positions now occur in `m`, `hmap` definitions will have to apply the higher-order function to these as well.
+
 - Adds `Functor` constraints to `hmap` and `Monad` constraints to `handle`, allowing a greater variety of instances to be defined (e.g. for recursively-nested syntax).
 
 - Replaces the default definitions of `hmap` and `handle` with derivations based on `Generic1` instead of `Coercible`. Therefore, first-order effects wishing to derive these instances will require `Generic1` instances, presumably derived using `-XDeriveGeneric`.
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -8,11 +8,11 @@
 
 - Changes the signature of `eff` from `sig m (m a) -> m a` to `sig m a -> m a`, requiring effects to hold `m k` in their continuation positions instead of merely `k`. This was done in order to improve interoperability with other presentations of higher-order syntax, e.g. `bound`; syntax used with `bound` can now be given `HFunctor` and `Carrier` instances.
 
-- Removes `fmap'`, as it is now obsolete.
-
 - Adds `Functor` constraints to `hmap` and `Monad` constraints to `handle`, allowing a greater variety of instances to be defined (e.g. for recursively-nested syntax).
 
 - Replaces the default definitions of `hmap` and `handle` with derivations based on `Generic1` instead of `Coercible`. Therefore, first-order effects wishing to derive these instances will require `Generic1` instances, presumably derived using `-XDeriveGeneric`.
+
+- Deprecates `fmap'` in favour of `fmap`.
 
 - Deprecates `handlePure` in favour of `hmap`.
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -10,6 +10,8 @@
 
 - Adds `Functor` constraints to `hmap` and `Monad` constraints to `handle`, allowing a greater variety of instances to be defined (e.g. for recursively-nested syntax).
 
+- Replaces the default definitions of `hmap` and `handle` with derivations based on `Generic1` instead of `Coercible`. Therefore, first-order effects wishing to derive these instances will require `Generic1` instances, presumably derived using `-XDeriveGeneric`.
+
 # v0.4.0.0
 
 ## Backwards-incompatible changes
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -12,6 +12,8 @@
 
 - Replaces the default definitions of `hmap` and `handle` with derivations based on `Generic1` instead of `Coercible`. Therefore, first-order effects wishing to derive these instances will require `Generic1` instances, presumably derived using `-XDeriveGeneric`.
 
+- Deprecates `handlePure` in favour of `hmap`.
+
 # v0.4.0.0
 
 ## Backwards-incompatible changes
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,12 @@
 
 - Replaces `runResource` with an equivalent function that uses `MonadUnliftIO` to select the correct unlifting function (a la `withResource`, which is removed in favor of `runResource`).
 
+- Changes the signature of `eff` from `sig m (m a) -> m a` to `sig m a -> m a`, requiring effects to hold `m k` in their continuation positions instead of merely `k`. This was done in order to improve interoperability with other presentations of higher-order syntax, e.g. `bound`; syntax used with `bound` can now be given `HFunctor` and `Carrier` instances.
+
+- Removes `fmap'`, as it is now obsolete.
+
+- Adds `Functor` constraints to `hmap` and `Monad` constraints to `handle`, allowing a greater variety of instances to be defined (e.g. for recursively-nested syntax).
+
 # v0.4.0.0
 
 ## Backwards-incompatible changes
--- a/docs/common_errors.md
+++ b/docs/common_errors.md
@@ -8,40 +8,38 @@ progress.)
 
 ## I'm getting kind errors when implementing a `Carrier` instance!
 
-Given a `Teletype` data type:
+Given an effect datatype that doesn’t use the `m` parameter:
 
 ```haskell
-data Teletype m k
-  = Read (String -> k)
-  | Write String k
+data Fail m k
+  = Fail String
   deriving (Functor)
 
-newtype TeletypeIOC m a = TeletypeIOC { runTeletypeIOC :: m a }
+newtype FailC m a = FailC { runFailC :: m (Either String a) }
 ```
 
 Declaring a `Carrier` instance will fail:
 
 ```haskell
 instance (Carrier sig m, Effect sig)
-    => Carrier (Teletype :+: sig) (TeletypeIOC m) where…
+    => Carrier (Fail :+: sig) (FailC m) where…
 ```
 
 ```
 • Expected kind ‘(* -> *) -> * -> *’,
-    but ‘Teletype :+: sig’ has kind ‘* -> * -> *’
-• In the first argument of ‘Carrier’, namely ‘(Teletype :+: sig)’
+    but ‘Fail :+: sig’ has kind ‘* -> * -> *’
+• In the first argument of ‘Carrier’, namely ‘(Fail :+: sig)’
   In the instance declaration for
-    ‘Carrier (Teletype :+: sig) (TeletypeIOC m)
+    ‘Carrier (Fail :+: sig) (FailC m)
 ```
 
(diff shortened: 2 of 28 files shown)


ChangeLog.md
```diff
@@ -1,9 +1,21 @@
 # v0.5.0.0
 
+- Derives `Generic1` instances for all non-existentially-quantified effect datatypes.
```
robrix: This isn’t precisely the same as “all first-order effect datatypes” because `Resumable` uses an existential and thus doesn’t admit a `Generic1` instance.

docs/common_errors.md
```diff
@@ -8,40 +8,38 @@ progress.)
 
 ## I'm getting kind errors when implementing a `Carrier` instance!
 
-Given a `Teletype` data type:
+Given an effect datatype that doesn’t use the `m` parameter:
```
robrix: `Teletype` now uses `m`, so the only effects falling into this category are ones which have no continuations whatsoever, e.g. `Fail`.

src/Control/Effect/Carrier.hs
```diff
@@ -1,31 +1,28 @@
...
+-- | Higher-order functors of kind @(* -> *) -> (* -> *)@ map functors to functors.
+--
+--   All effects must be 'HFunctor's.
 class HFunctor h where
-  -- | Functor map. This is required to be 'fmap'.
-  --
-  --   This can go away once we have quantified constraints.
-  fmap' :: (a -> b) -> (h m a -> h m b)
-  default fmap' :: Functor (h m) => (a -> b) -> (h m a -> h m b)
-  fmap' = fmap
```
robrix: 🔥

src/Control/Effect/Carrier.hs
```diff
@@ -34,46 +31,109 @@ class HFunctor h where
...
+  ghmap _ v = case v of {}
+
+instance GHFunctor m m' U1 U1 where
+  ghmap _ = id
+
+instance GHFunctor m m' (K1 R c) (K1 R c) where
+  ghmap _ = coerce
+
+instance (Functor f, GHFunctor m m' g g') => GHFunctor m m' (f :.: g) (f :.: g') where
+  ghmap f = Comp1 . fmap (ghmap f) . unComp1
```
robrix: There are other configurations possible here that this instance doesn’t cover, e.g. `m (f a)`, `m (m a)`, `m (f (m a))`, etc., but I wanted to avoid getting into overlapping instances.

src/Control/Effect/Carrier.hs
```diff
@@ -34,46 +31,109 @@ class HFunctor h where
...
+          -> rep' (f a)
+
+instance GEffect m m' rep rep' => GEffect m m' (M1 i c rep) (M1 i c rep') where
+  ghandle state handler = M1 . ghandle state handler . unM1
+
+instance (GEffect m m' l l', GEffect m m' r r') => GEffect m m' (l :+: r) (l' :+: r') where
+  ghandle state handler (L1 l) = L1 (ghandle state handler l)
+  ghandle state handler (R1 r) = R1 (ghandle state handler r)
+
+instance (GEffect m m' l l', GEffect m m' r r') => GEffect m m' (l :*: r) (l' :*: r') where
```
robrix: This does not implement left-to-right threading of state through multiple higher-order positions, but that really only occurs when you’ve got existentials, which can’t use the generic derivations anyway.

src/Control/Effect/Carrier.hs
```diff
@@ -1,31 +1,28 @@
...
-  -- A definition for 'hmap' over first-order effects can be derived automatically.
-  hmap :: (forall x . m x -> n x) -> (h m a -> h n a)
-
-  default hmap :: Coercible (h m a) (h n a)
-               => (forall x . m x -> n x)
-               -> (h m a -> h n a)
-  hmap _ = coerce
+  --
+  -- A definition for 'hmap' over first-order effects can be derived automatically provided a 'Generic1' instance is available.
+  hmap :: Functor m => (forall x . m x -> n x) -> (h m a -> h n a)
```
robrix: The addition of a `Functor` constraint for `m` means that we admit strictly more instances, e.g.:

```haskell
newtype Scope f a = Scope (f (Maybe (f a)))

instance HFunctor Scope where hmap f (Scope b) = Scope (f (fmap (fmap f) b))
```

would not be possible without the `Functor` constraint on `m`.

On the other hand, it also means that we allow strictly fewer callers, since before you could’ve `hmap`’d to shove a `Set` in there.

I’m ok with this. Use _The Constrained-Monad Problem_ or `Coyoneda` if you want to do that sort of thing.

src/Control/Effect/Carrier.hs
```diff
@@ -34,46 +31,109 @@ class HFunctor h where
 --   1. Be functorial in their last two arguments, and
 --   2. Support threading effects in higher-order positions through using the carrier’s suspended state.
 --
--- All first-order effects (those without recursive occurrences of @m@) admit a default definition
--- of 'handle'. The @-XDeriveAnyClass@ extension allows derivation of both 'HFunctor' and 'Effect':
```
robrix: I removed this part of the note because `DeriveAnyClass` is only necessary to name the classes in `deriving` clauses, not to actually derive the instance. E.g. this with `-XDeriveAnyClass`:

```haskell
data State s m k
  = Get (s -> m k)
  | Put s (m k)
  deriving (Functor, Generic1, HFunctor, Effect)
```

is equivalent to this without:

```haskell
data State s m k
  = Get (s -> m k)
  | Put s (m k)
  deriving (Functor, Generic1)

instance HFunctor (State s)
instance Effect (State s)
```

Users don’t have to do anything special to use the latter, so I didn’t feel it was worth calling out in such a prominent place.

src/Control/Effect/Carrier.hs
```diff
@@ -34,46 +31,109 @@ class HFunctor h where
...
+  handle state handler = to1 . ghandle state handler . from1
   {-# INLINE handle #-}
 
 
-
 -- | The class of carriers (results) for algebras (effect handlers) over signatures (effects), whose actions are given by the 'eff' method.
 class (HFunctor sig, Monad m) => Carrier sig m | m -> sig where
   -- | Construct a value in the carrier for an effect signature (typically a sum of a handled effect and any remaining effects).
-  eff :: sig m (m a) -> m a
+  eff :: sig m a -> m a
```
robrix: Finn the Human from Adventure Time pumping his fist in the air and shouting “algebraic!”

src/Control/Effect/Resumable.hs
```diff
@@ -22,12 +22,16 @@ import Control.Monad.Trans.Class
...
-deriving instance Functor (Resumable err m)
-deriving instance HFunctor (Resumable err)
-deriving instance Effect (Resumable err)
+deriving instance Functor m => Functor (Resumable err m)
+
+instance HFunctor (Resumable err) where
+  hmap f (Resumable err k) = Resumable err (f . k)
+
+instance Effect (Resumable err) where
+  handle state handler (Resumable err k) = Resumable err (handler . (<$ state) . k)
```
robrix: We can no longer derive these two instances, because deriving them requires a `Generic1` instance, which itself can’t be derived for `Resumable` due to the existential quantification.

ChangeLog.md
```diff
@@ -1,9 +1,21 @@
...
+
 ## Backwards-incompatible changes
 
 - Replaces `runResource` with an equivalent function that uses `MonadUnliftIO` to select the correct unlifting function (a la `withResource`, which is removed in favor of `runResource`).
 
+- Changes the signature of `eff` from `sig m (m a) -> m a` to `sig m a -> m a`, requiring effects to hold `m k` in their continuation positions instead of merely `k`. This was done in order to improve interoperability with other presentations of higher-order syntax, e.g. `bound`; syntax used with `bound` can now be given `HFunctor` and `Carrier` instances.
```
patrickt: Can we add a short blurb here on how to upgrade? e.g. 

> To upgrade from versions of `fused-effects` < 0.4, convert a handler’s continuations to be `m k` rather than `k`. If no existential type variables appear in your effect definition, you can derive an `HFunctor` and `Effect` instance. Otherwise, implement the required instances by hand.

docs/common_errors.md
patrickt: Honestly now that almost every effect uses its `m`, we could probably nuke this whole documentation.

src/Control/Effect/Carrier.hs
```diff
@@ -1,79 +1,151 @@
...
 --   This is applicable whenever @f@ is 'Coercible' to @g@, e.g. simple @newtype@s.
-handleCoercible :: (HFunctor sig, Coercible f g) => sig f (f a) -> sig g (g a)
-handleCoercible = handlePure coerce
+handleCoercible :: (HFunctor sig, Functor f, Coercible f g) => sig f a -> sig g a
+handleCoercible = hmap coerce
 {-# INLINE handleCoercible #-}
+
+
+-- | Generic implementation of 'HFunctor'.
+class GHFunctor m m' rep rep' where
```
patrickt: Do we need to INLINE the definitions here? Can we check the generated Core/add to the inspection tests to make sure we’re not missing these opportunities?

src/Control/Effect/Carrier.hs
patrickt: I think this is fair. More delicate instances can be covered by `fused-effects-th`, when it’s done.

src/Control/Effect/Carrier.hs
```diff
@@ -1,79 +1,151 @@
...
+
+instance (GEffect m m' l l', GEffect m m' r r') => GEffect m m' (l :+: r) (l' :+: r') where
+  ghandle state handler (L1 l) = L1 (ghandle state handler l)
+  ghandle state handler (R1 r) = R1 (ghandle state handler r)
+
+instance (GEffect m m' l l', GEffect m m' r r') => GEffect m m' (l :*: r) (l' :*: r') where
+  ghandle state handler (l :*: r) = ghandle state handler l :*: ghandle state handler r
+
+instance GEffect m m' V1 V1 where
+  ghandle _ _ v = case v of {}
```
patrickt: Zen.

src/Control/Effect/Pure.hs
```diff
@@ -8,18 +8,11 @@ module Control.Effect.Pure
 import Control.Applicative
 import Control.Effect.Carrier
 import Data.Coerce
+import GHC.Generics (Generic1)
```
patrickt: Do we want to export `Generic1` from `Control.Effect.Sum` or something?

src/Control/Effect/Resumable.hs
```diff
@@ -82,7 +86,7 @@ runResumable :: ResumableC err m a -> m (Either (SomeError err) a)
 runResumable = runError . runResumableC
 
 newtype ResumableC err m a = ResumableC { runResumableC :: ErrorC (SomeError err) m a }
-  deriving newtype (Alternative, Applicative, Functor, Monad, MonadFail, MonadIO, MonadPlus, MonadTrans)
```
patrickt: I’d prefer the explicit deriving strategy. Leaving it tacit can introduce surprising bugs and helps me keep less in my head.

src/Control/Effect/Sum.hs
```diff
@@ -1,30 +1,21 @@
...
 ) where
 
 import Control.Effect.Carrier
+import GHC.Generics (Generic1)
 
 data (f :+: g) (m :: * -> *) k
   = L (f m k)
   | R (g m k)
-  deriving (Eq, Functor, Ord, Show)
+  deriving stock (Eq, Functor, Generic1, Ord, Show)
```
patrickt: Love that we can just derive this now.

src/Control/Effect/State/Internal.hs
```diff
@@ -1,4 +1,4 @@
-{-# LANGUAGE DeriveAnyClass, DeriveFunctor, ExplicitForAll, FlexibleContexts, FlexibleInstances, KindSignatures, MultiParamTypeClasses, TypeOperators, UndecidableInstances #-}
+{-# LANGUAGE DeriveAnyClass, DeriveFunctor, DeriveGeneric, DerivingStrategies, ExplicitForAll, FlexibleContexts, FlexibleInstances, MultiParamTypeClasses, TypeOperators, UndecidableInstances #-}
```
patrickt: Can we add these extensions to the section in the README that describes the commonly-used extensions?

docs/common_errors.md
robrix: I had considered that, but `Fail` doesn’t, and it’s conceivable that you might write your own fail-alike effect, e.g. #154.

src/Control/Effect/Pure.hs
robrix: I don’t think so. It’s only needed about half the time, and it’s in `base`.

src/Control/Effect/Resumable.hs
robrix: Fair, tho I’d note we do have warnings for those turned on.

PR #188, opened 2019-07-06
MonadFix instances

robrix:
This PR defines `MonadFix` instances for all the carriers.

- [x] Depends on #184.

### DIFF
--- a/src/Control/Effect/Error.hs
+++ b/src/Control/Effect/Error.hs
@@ -12,6 +12,7 @@ import Control.Effect.Carrier
 import Control.Effect.Sum
 import Control.Monad (MonadPlus(..), (<=<))
 import Control.Monad.Fail
+import Control.Monad.Fix
 import Control.Monad.IO.Class
 import Control.Monad.Trans.Class
 import Prelude hiding (fail)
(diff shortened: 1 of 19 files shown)


src/Control/Effect/Error.hs
```diff
@@ -76,6 +77,10 @@ instance Monad m => Monad (ErrorC e m) where
   ErrorC a >>= f = ErrorC (a >>= either (pure . Left) (runError . f))
   {-# INLINE (>>=) #-}
 
+instance MonadFix m => MonadFix (ErrorC e m) where
+  mfix f = ErrorC (mfix (runError . either (error "mfix (ErrorC): function returned failure") f))
```
robrix: Like `ExceptT`, this is safe to use provided that you ensure that throwing an error means you won’t inspect the argument.

src/Control/Effect/Interpose.hs
```diff
@@ -27,7 +32,7 @@ runInterpose :: (forall x . eff m x -> m x) -> InterposeC eff m a -> m a
 runInterpose handler = runReader (Handler handler) . runInterposeC
 
 newtype InterposeC eff m a = InterposeC { runInterposeC :: ReaderC (Handler eff m) m a }
-  deriving (Applicative, Functor, Monad)
+  deriving (Alternative, Applicative, Functor, Monad, MonadFail, MonadFix, MonadIO, MonadPlus)
```
robrix: Added a few other missing instances while I was at it.

src/Control/Effect/NonDet.hs
```diff
@@ -61,6 +62,10 @@ instance MonadFail m => MonadFail (NonDetC m) where
   fail s = NonDetC (\ _ _ -> fail s)
   {-# INLINE fail #-}
 
+instance MonadFix m => MonadFix (NonDetC m) where
+  mfix f = NonDetC (\ cons nil -> mfix (\ a -> runNonDetC (f (head a)) (fmap . (:)) (pure [])) >>= foldr cons nil)
```
robrix: This one took a little ingenuity, let me tell you.

PR #189, opened 2019-07-13
Improve haddocks a bit

mitchellwrosen:
This patch adds doc headers to separate effects from carrier(s). I also put re-exports at the bottom, so they are more out of the way (`Control.Effect.NonDet` for example currently has a large re-export of `Alternative` before listing carriers).


src/Control/Effect/Fail.hs
```diff
@@ -1,9 +1,12 @@
...
 module Control.Effect.Fail
-( Fail(..)
-, MonadFail(..)
+( -- * Fail effect
+  Fail(..)
+  -- * Fail carrier
 , runFail
 , FailC(..)
+  -- * Re-exports
+, MonadFail(..)
```
robrix: I really like separating re-exports like this 👍

robrix:
Build failed due to flakiness, retrying.

PR #190, opened 2019-07-13
Derive Foldable & Traversable instance for Sum

robrix:
Does what it says on the tin ☝️


src/Control/Effect/Sum.hs
```diff
@@ -11,7 +11,7 @@ import GHC.Generics (Generic1)
 data (f :+: g) (m :: * -> *) k
   = L (f m k)
   | R (g m k)
-  deriving stock (Eq, Functor, Generic1, Ord, Show)
+  deriving stock (Eq, Foldable, Functor, Generic1, Ord, Show, Traversable)
```
robrix: Syntax frequently wants `Foldable` and `Traversable` instances for e.g. collecting or renaming free variables.

PR #195, opened 2019-08-09
Actions-based CI

robrix:
Travis is slow and flaky; let’s try to be less slow and flaky!


robrix:
Even without caching this is now competitive with travis thanks to @tclem’s excellent work on `setup-haskell` ✨

robrix:
@patrickt: This is in the ballpark of travis CI now, so I’d like to see about merging this and then adding caching once that’s available.

.github/workflows/haskell.yml
```diff
@@ -0,0 +1,37 @@
...
+
+    - uses: actions/setup-haskell@v1
+      name: Setup Haskell
+      with:
+        ghc-version: ${{ matrix.ghc }}
+        cabal-version: ${{ matrix.cabal }}
+
+    - name: Install dependencies
+      run: |
+        cabal v2-update
```
robrix: I’m not passing `-v` because I don’t think the extra noise is all that useful.

robrix:
I’mma change the branch protections over and then merge when this goes green.

PR #197, opened 2019-08-25
Binary nondeterminism

robrix:
This PR redefines `NonDetC`, changing it from a Church-encoded list to a Church-encoded binary tree. This should result in greater preservation of parallelism between branches, tho I haven’t actually measured that. It should also result in slightly simpler Core since e.g. `pure` no longer needs to invoke both continuations, but again, I haven’t measured that.


src/Control/Effect/Cull.hs
```diff
@@ -51,18 +52,20 @@ cull m = send (Cull m pure)
...
   empty = CullC empty
   {-# INLINE empty #-}
-  l <|> r = CullC $ ReaderC $ \ cull -> NonDetC $ \ cons nil -> do
-    runNonDetC (runReader cull (runCullC l))
-      (\ a as -> cons a (if cull then nil else as))
-      (runNonDetC (runReader cull (runCullC r)) cons nil)
+  l <|> r = CullC $ ReaderC $ \ cull ->
+    if cull then
+      NonDetC $ \ fork leaf nil ->
+        runNonDetC (runReader cull (runCullC l)) fork leaf (runNonDetC (runReader cull (runCullC r)) fork leaf nil)
```
robrix: When we’re inside a `cull`, we alternate by using the rhs as the `nil` continuation for the lhs.

src/Control/Effect/Cull.hs
```diff
@@ -51,18 +52,20 @@ cull m = send (Cull m pure)
...
+    else
+      runReader cull (runCullC l) <|> runReader cull (runCullC r)
```
robrix: When we aren’t inside a `cull`, then we just alternate as normal for the underlying `NonDetC`.

src/Control/Effect/Cut.hs
```diff
@@ -102,15 +102,15 @@ instance Monad (CutC m) where
...
+  fail s = lift (Fail.fail s)
   {-# INLINE fail #-}
 
 instance MonadFix m => MonadFix (CutC m) where
   mfix f = CutC (\ cons nil _ -> mfix (\ a -> runCutC (f (head a)) (fmap . (:)) (pure []) (pure [])) >>= foldr cons nil)
   {-# INLINE mfix #-}
 
 instance MonadIO m => MonadIO (CutC m) where
-  liftIO io = CutC (\ cons nil _ -> liftIO io >>= flip cons nil)
+  liftIO io = lift (liftIO io)
```
robrix: I attempted to apply the same approach to `CutC`, but (much to my surprise) the semantics of `cut`/`call` are such that you actually _want_ the left-to-right sequencing of the encoded list formulation. I’m not yet convinced that it _can’t_ be done with a binary tree formulation, but it sure isn’t straightforward.

src/Control/Effect/NonDet.hs
```diff
@@ -12,81 +12,121 @@ module Control.Effect.NonDet
...
 
 data NonDet m k
   = Empty
   | Choose (Bool -> m k)
-  deriving stock (Functor, Generic1)
-  deriving anyclass (HFunctor, Effect)
+  deriving (Functor, Generic1)
+
+instance HFunctor NonDet
+instance Effect   NonDet
```
robrix: I honestly don’t think we should use `DeriveAnyClass` for these, as it’s a false economy.

src/Control/Effect/NonDet.hs
```diff
@@ -12,81 +12,121 @@ module Control.Effect.NonDet
...
 
 
 -- | Run a 'NonDet' effect, collecting all branches’ results into an 'Alternative' functor.
 --
---   Using '[]' as the 'Alternative' functor will produce all results, while 'Maybe' will return only the first. However, unlike 'runNonDetOnce', this will still enumerate the entire search space before returning, meaning that it will diverge for infinite search spaces, even when using 'Maybe'.
+--   Using @[]@ as the 'Alternative' functor will produce all results, while 'Maybe' will return only the first. However, unlike 'Control.Effect.Cull.runNonDetOnce', this will still enumerate the entire search space before returning, meaning that it will diverge for infinite search spaces, even when using 'Maybe'.
```
robrix: Haddock can’t link to `[]` because `[]` isn’t defined in (Haskell) source. Syntax sugar!

src/Control/Effect/NonDet.hs
```diff
@@ -12,81 +12,121 @@ module Control.Effect.NonDet
...
 runNonDet :: (Alternative f, Applicative m) => NonDetC m a -> m (f a)
-runNonDet (NonDetC m) = m (fmap . (<|>) . pure) (pure empty)
+runNonDet (NonDetC m) = m (liftA2 (<|>)) (pure . pure) (pure empty)
 
 -- | A carrier for 'NonDet' effects based on Ralf Hinze’s design described in [Deriving Backtracking Monad Transformers](https://www.cs.ox.ac.uk/ralf.hinze/publications/#P12).
 newtype NonDetC m a = NonDetC
-  { -- | A higher-order function receiving two parameters: a function to combine each solution with the rest of the solutions, and an action to run when no results are produced.
-    runNonDetC :: forall b . (a -> m b -> m b) -> m b -> m b
+  { -- | A higher-order function receiving three continuations, respectively implementing '<|>', 'pure', and 'empty'.
+    runNonDetC :: forall b . (m b -> m b -> m b) -> (a -> m b) -> m b -> m b
```
robrix: This is so much cleaner ✨

src/Control/Effect/NonDet.hs
```diff
@@ -12,81 +12,121 @@ module Control.Effect.NonDet
...
   }
-  deriving stock (Functor)
+  deriving (Functor)
 
+-- $
+--   prop> run (runNonDet (pure a *> pure b)) === Just b
+--   prop> run (runNonDet (pure a <* pure b)) === Just a
```
robrix: These sorts of tests of lawfulness strike me as something we could use more of.

src/Control/Effect/NonDet.hs
```diff
@@ -12,81 +12,121 @@ module Control.Effect.NonDet
...
   {-# INLINE fail #-}
 
 instance MonadFix m => MonadFix (NonDetC m) where
-  mfix f = NonDetC (\ cons nil -> mfix (\ a -> runNonDetC (f (head a)) (fmap . (:)) (pure [])) >>= foldr cons nil)
+  mfix f = NonDetC $ \ fork leaf nil ->
+    mfix (\ a -> runNonDetC (f (fromJust (fold (<|>) Just Nothing a)))
+      (liftA2 Fork)
+      (pure . Leaf)
+      (pure Nil))
+    >>= fold fork leaf nil
```
robrix: Note that this implementation has the same behaviour as the previous one, and as such is subject to #209. I do not intend to fix that behaviour in this PR.

src/Control/Effect/NonDet.hs
```diff
@@ -12,81 +12,121 @@ module Control.Effect.NonDet
...
 instance MonadIO m => MonadIO (NonDetC m) where
-  liftIO io = NonDetC (\ cons nil -> liftIO io >>= flip cons nil)
+  liftIO io = lift (liftIO io)
   {-# INLINE liftIO #-}
 
 instance MonadPlus (NonDetC m)
 
 instance MonadTrans NonDetC where
-  lift m = NonDetC (\ cons nil -> m >>= flip cons nil)
+  lift m = NonDetC (\ _ leaf _ -> m >>= leaf)
```
robrix: We no longer invoke the `nil` continuation for `pure` & `lift`, which should make this slightly more efficient in general.

src/Control/Effect/NonDet.hs
```diff
@@ -12,81 +12,121 @@ module Control.Effect.NonDet
...
+
+instance Applicative BinaryTree where
+  pure = Leaf
+  Nil      <*> _ = Nil
+  Leaf f   <*> a = fmap f a
+  Fork a b <*> c = Fork (a <*> c) (b <*> c)
+
+instance Alternative BinaryTree where
+  empty = Nil
+  (<|>) = Fork
```
robrix: We need the `Alternative` instance because `runNonDet` (used in `eff`) requires it. cf #207.

src/Control/Effect/NonDet.hs
```diff
@@ -12,81 +12,121 @@ module Control.Effect.NonDet
...
 instance (Carrier sig m, Effect sig) => Carrier (NonDet :+: sig) (NonDetC m) where
   eff (L Empty)      = empty
   eff (L (Choose k)) = k True <|> k False
-  eff (R other)      = NonDetC $ \ cons nil -> eff (handle [()] (fmap concat . traverse runNonDet) other) >>= foldr cons nil
+  eff (R other)      = NonDetC $ \ fork leaf nil -> eff (handle (Leaf ()) (fmap join . traverse runNonDet) other) >>= fold fork leaf nil
   {-# INLINE eff #-}
 
 
+data BinaryTree a = Nil | Leaf a | Fork (BinaryTree a) (BinaryTree a)
+  deriving (Eq, Foldable, Functor, Ord, Show, Traversable)
```
robrix: I’ve opted not to export this for the time being, but I feel a bit weird about that. I don’t know that I really want to add a `Data.BinaryTree` module, either, since that seems like a potentially contended name for the module. Tho perhaps I worry too much, since hoogle doesn’t seem to know about any such modules?

src/Control/Effect/NonDet.hs
```diff
@@ -12,81 +12,117 @@ module Control.Effect.NonDet
...
+
+instance Applicative BinaryTree where
+  pure = Leaf
+  f <*> a = fold Fork (<$> a) Nil f
```
robrix: This and the definition of `>>=` make me extremely happy.

src/Control/Effect/NonDet.hs
```diff
@@ -12,81 +12,117 @@ module Control.Effect.NonDet
...
+
+instance Alternative BinaryTree where
+  empty = Nil
+  (<|>) = Fork
+
+instance Monad BinaryTree where
+  a >>= f = fold Fork f Nil a
```
robrix: `>>=` on trees is substitution.

src/Control/Effect/NonDet.hs
patrickt: I don’t object to keeping it hidden, personally. The more details we hide, the more we can iterate on the internals without breaking public API. We’re already entailing a good bit of churn these days, after all!

patrickt:
Informal first observations based off of #211 indicate this gives a slight speedup, which is further improved by marking the BinaryTree Functor/Monad/etc. methods as INLINE. ✨

robrix:
That’s super exciting!

src/Control/Effect/Cull.hs
```diff
@@ -51,18 +52,20 @@ cull m = send (Cull m pure)
 --
 --   prop> run (runNonDet (runCull (pure a <|> pure b))) === [a, b]
 runCull :: Alternative m => CullC m a -> m a
-runCull (CullC m) = runNonDetC (runReader False m) ((<|>) . pure) empty
+runCull (CullC m) = runNonDetC (runReader False m) (<|>) pure empty
```
patrickt: I like that this got simpler.

src/Control/Effect/NonDet.hs
patrickt: Up to you. I personally like it, but given that it’s a two-line difference I’m not worked up about it.

src/Control/Effect/NonDet.hs
```diff
@@ -12,81 +12,117 @@ module Control.Effect.NonDet
...
-  eff (R other)      = NonDetC $ \ cons nil -> eff (handle [()] (fmap concat . traverse runNonDet) other) >>= foldr cons nil
+  eff (R other)      = NonDetC $ \ fork leaf nil -> eff (handle (Leaf ()) (fmap join . traverse runNonDet) other) >>= fold fork leaf nil
   {-# INLINE eff #-}
 
 
+data BinaryTree a = Nil | Leaf a | Fork (BinaryTree a) (BinaryTree a)
+  deriving (Eq, Foldable, Functor, Ord, Show, Traversable)
+
+instance Applicative BinaryTree where
+  pure = Leaf
```
patrickt: Can you `INLINE` the various instances on `BinaryTree`?

src/Control/Effect/Cut.hs
patrickt: Can you walk me through what’s going on here? I’m not sure I follow the operative difference introduced by this change.

src/Control/Effect/NonDet.hs
robrix: That is an excellent point.

src/Control/Effect/Cut.hs
robrix: The semantics are unchanged, because:

```haskell
lift m = CutC (\ cons nil _ -> m >>= flip cons nil)
```

thus,

```haskell
lift (liftIO io) = CutC (\ cons nil _ -> liftIO io >>= flip cons nil)
```

which is exactly the definition we’ve replaced.

robrix:
hypnotoad on ice

robrix:
> I would mark the changelog entry for this patch as backwards-compatible.

The type of `NonDetC` & `runNonDetC` changed, and both are exported, so technically it’s not, alas.

Issue #183, opened 2019-07-05
Define (terminal) Carrier instances for common monads?


robrix:
E.g.:

```haskell
instance Carrier (Reader r) ((->) r) where …
instance Carrier (Error e) (Either e) where …
instance Carrier NonDet [] where …
instance Carrier NonDet Maybe where …
```

**Pro:**

These would make it _extremely_ convenient to e.g. use `Error` effects in pure code.

**Con:**

We can only define one `Carrier` instance per monad, so we couldn’t also define e.g. `Carrier Fail (Either String)`. This could limit the usefulness of these instances?

**Alt:**

1. We could define tiny `newtype` helpers for these tasks, e.g.:

    ```haskell
    newtype EitherC e a = EitherC { runEitherC :: Either e a }
      deriving …
    instance Carrier (Error e) (EitherC e) where …
    
    foo :: Either A B
    foo = runEitherC $ … `catchError` \ e -> …
    ```

    These would give us essentially the same interface, just with the `newtype` wrapper around the result monad.

2. We could define simple wrappers using our extant carriers, e.g.:

    ```haskell
    runEither :: ErrorC e PureC a -> Either e a
    runEither = run . runError
    ```

    AFAICS this would be equivalent to the above and should still let us `coerce` the carrier all the way to the result monad, i.e. we should also be able to define `runEither = coerce`.

robrix:
**NB:** The second alternative would work for `NonDetC`, but would _not_ allow us to define it by coercion, whereas the following would:

```haskell
newtype ListC a = ListC { runListC :: [a] }
  deriving …

instance Carrier NonDet ListC where …

newtype MaybeC a = MaybeC { runMaybeC :: Maybe a }
  deriving …

instance Carrier NonDet MaybeC where …
```

Issue #142, opened 2019-03-28
Effect laws are undocumented


robrix:
Effects typically have laws which carriers should observe. e.g., given a `Reader` effect, sequential `ask`s in the same `do`-block are expected to return the same value (whereas a carrier for e.g. `Reader Int` could be implemented to store and increment the value before returning). Unfortunately, these are currently undocumented, which makes it harder to know that your carrier is (in addition to type-checking) correct.

**NB:** Some of these might be somewhat subtle when considering interactions between effects; e.g. `State` inside `NonDet` is maintained per-branch, whereas `State` outside `NonDet` is global. This is a feature, in general, since it allows you to select the desired composite semantics by reordering, but it may make it harder to state some of the laws succinctly.

For example, it’s tempting to say that `ask >>= \ before -> m >> ask >>= \ after -> pure (before == after)` should (observationally) equal `pure True` for arbitrary `m`, but if `m` doesn’t continue (e.g. because it contains a `throwError`, `fail`, or `empty`), that is not strictly true.

PR #204, opened 2019-09-01
Separate carrier and effect modules

robrix:
This PR separates carriers and effects into disjoint hierarchies. Goals include:

- More clearly delineating interfaces and implementations: interfaces go in `Control.Effect.*`, implementations in `Control.Carrier.*`.
- Easier experimentation with new carriers: there’s an easy and obvious place to put new carrier modules; carriers could potentially be defined for monads from `base` & `transformers`. (cf #183) I intend to explore this in separate PRs.
- Consistent API between carriers. We could potentially even abstract carriers via typeclasses or Backpack signatures. Even absent signatures, sharing the same interface between multiple carriers also implies that swapping one for another can be as simple as switching which module you import.
- Making the choice of which carrier to use conscious & explicit: “I imported strict `StateC`” instead of “whichever one the effect module re-exports.”
- Retaining good import ergonomics: carrier modules re-export the effect & `Carrier` machinery.
- Documenting the laws for each effect constructor, and then verifying them separately against each carrier (to be fully realized in a follow-up PR). (cf #142)

On the other hand, this is also quite a disruptive PR, since it means pretty significant changes for consumers of the library. I’d therefore like to get input from users on this!

It also builds on several other PRs currently slated for 0.6:

- [x] Depends on #196.
- [x] Depends on #197.
- [x] Depends on #198.
- [x] Depends on #199.
- [x] Depends on #200.
- [x] Depends on #205.
- [x] ~~Add `Control.Effect.Carrier` back in with a deprecation notice.~~ We’ll do this if we ship this as 0.6, but we might just ship it + some other stuff as 1.0; so will tackle that separately.
- [x] Add `Control.Effect.NonDet` back in, re-exporting `Choose` & `Empty`.


robrix:
Thoughts on renaming `Control.Carrier.Empty.Maybe` and `Control.Carrier.Either.Error` to `.Direct`, in keeping with the `.Church` names for the various church-encoded nondeterminism carriers?

robrix:
Reasonably representative example of migrating a dependent project over: https://github.com/robrix/path/pull/109/files

robrix:
cc @jkachmar — I’d specifically love to hear your thoughts about this change!

robrix:
> Thoughts on renaming `Control.Carrier.Empty.Maybe` and `Control.Carrier.Either.Error` to `.Direct`, in keeping with the `.Church` names for the various church-encoded nondeterminism carriers?

Answering my own question: we might want to offer a selection of direct-style carriers, e.g. a `Fail` carrier which always ignores the string and returns in `Maybe`; so I think we want to keep naming them for the type produced.

src/Control/Effect/Choose.hs
```diff
@@ -52,69 +48,11 @@ some1 :: (Carrier sig m, Member Choose sig) => m a -> m (NonEmpty a)
...
-  {-# INLINE pure #-}
-  f <*> a = fold Fork (<$> a) f
-  {-# INLINE (<*>) #-}
-
-instance Monad BinaryTree where
-  a >>= f = fold Fork f a
-  {-# INLINE (>>=) #-}
+newtype Choosing m a = Choosing { getChoosing :: m a }
 
+instance (Carrier sig m, Member Choose sig) => S.Semigroup (Choosing m a) where
```
robrix: Kinda snuck this in here, but it’s extremely convenient to have available. We’re using an explicit `qualified Data.Semigroup as S` import here to accommodate ghc 8.2.

src/Control/Effect/Empty.hs
```diff
@@ -32,63 +20,17 @@ instance Effect   Empty
...
+empty = send Empty
 
-instance (Carrier sig m, Effect sig) => Carrier (Empty :+: sig) (EmptyC m) where
-  eff (L Empty) = EmptyC (pure Nothing)
-  eff (R other) = EmptyC (eff (handle (Just ()) (maybe (pure Nothing) runEmptyC) other))
-  {-# INLINE eff #-}
+-- | Conditional failure, returning only if the condition is 'True'.
+guard :: (Carrier sig m, Member Empty sig) => Bool -> m ()
+guard True  = pure ()
+guard False = empty
```
robrix: Snuck this in as well, with essentially the same rationale as above.

src/Control/Effect/NonDet.hs
```diff
@@ -45,99 +28,16 @@ runNonDet (NonDetC m) = m (liftA2 (<|>)) (pure . pure) (pure empty)
...
-
-instance MonadIO m => MonadIO (NonDetC m) where
-  liftIO io = lift (liftIO io)
-  {-# INLINE liftIO #-}
+oneOf = foldMapA pure
 
-instance MonadPlus (NonDetC m)
+-- | Map a 'Foldable' collection of values into a nondeterministic computation using the supplied action.
+foldMapA :: (Foldable t, Alternative m) => (a -> m b) -> t a -> m b
+foldMapA f = getAlt #. foldMap (Alt #. f)
```
robrix: Snuck this in too. I end up defining this in just about every project that I use `Alternative` in; I’m considering proposing its addition to `base` (tho prolly under a different name, maybe `asumMap` because `concat` : `concatMap` :: `asum` : `asumMap`).

src/Control/Effect/NonDet.hs
```diff
@@ -45,99 +28,16 @@ runNonDet (NonDetC m) = m (liftA2 (<|>)) (pure . pure) (pure empty)
...
--- $setup
--- >>> :seti -XFlexibleContexts
--- >>> import Test.QuickCheck
--- >>> import Control.Effect.Pure
--- >>> import Data.Foldable (asum)
+-- | Compose a function operationally equivalent to 'id' on the left.
+--
+--   cf https://github.com/fused-effects/diffused-effects/pull/1#discussion_r323560758
+(#.) :: Coercible b c => (b -> c) -> (a -> b) -> (a -> c)
+(#.) _ = coerce
```
robrix: 🎩 @patrickt for pointing this out

src/Control/Effect/Sum.hs
```diff
@@ -30,3 +37,9 @@ instance {-# OVERLAPPABLE #-} Member sub sup => Member sub (sub' :+: sup) where
   inj = R . inj
   prj (R g) = prj g
   prj _     = Nothing
+
+
+-- | Construct a request for an effect to be interpreted by some handler later on.
+send :: (Member effect sig, Carrier sig m) => effect m a -> m a
+send = eff . inj
+{-# INLINE send #-}
```
robrix: Glad to have this, and the `HFunctor`/`Effect` instances, back where they belong.

robrix:
🚀

PR #217, opened 2019-09-23
Simpler constraints

### LINKED ISSUE #212: Requiring both Carrier & Member constraints is inconvenient
Having to have both `Carrier` and `Member` constraints is an unnecessary wrinkle. @lexi-lambda pointed out some possibilities for unifying them—or at least, providing a unified interface—involving using a type family for a `Carrier`’s signature type instead of a functional dependency.

robrix:
This PR:

- [x] Splits `Member` into separate `Inject` and `Project` classes, retaining the latter solely for interposition. This makes both into single-method classes, whose dictionaries are essentially passed around as `newtype`s, removing extra indirection around the method. I haven’t noticed any particular speedup from this change, however; `send`/`inj` is not a bottleneck as far as I can tell.
- [x] Defines a `Has` constraint synonym as the combination of `Inject` and `Carrier`, and uses it for all effect smart constructors.
- [x] Exports `Has` from all carrier modules, replacing the exports of `Carrier` and `Member`.
- [x] Applies all of the above to the docs.
- [x] Export `Carrier` again; it’s useful when handling effects on top of some other effectful context.
- [x] Fixes #212.

Note that in contrast to the discussion in #212, this does not involve any changes to `Carrier` (i.e. using a type family instead of a functional dependency to provide the signature). Unfortunately, experimentation in [`diffused-effects`](https://github.com/fused-effects/diffused-effects) demonstrated that while the type families approach has better ergonomics (no `FlexibleInstances` or `UndecidableInstances` extensions required for most carrier instances, and no `sig` parameter floating around purposelessly in `Has` constraints), it has unacceptable consequences for compile-times.

Specifically, it triggers such a severe [coercion pile-up](https://gitlab.haskell.org/ghc/ghc/issues/8095) that having stacks as few as fifteen effects deep become completely unusable. (I left a stack of about 20 carriers going for about fifteen minutes before finally giving up and killing the `ghc` process.) This is an entirely normal sort of quantity for projects like `semantic` designed around effects, making it a show-stopper.

We can perhaps revisit that part of the equation later on, once the coercion pile-up bug is fixed (or rather, once it has been fixed for a few releases of `ghc`).

🎩 @lexi-lambda for the suggestion!

### DIFF
--- a/src/Control/Carrier.hs
+++ b/src/Control/Carrier.hs
@@ -4,9 +4,15 @@ module Control.Carrier
 , module Control.Carrier.Pure
 , module Control.Effect.Class
 , module Control.Effect.Sum
+, send
 ) where
 
 import Control.Carrier.Class
 import Control.Carrier.Pure
 import Control.Effect.Class
 import Control.Effect.Sum
+
+-- | Construct a request for an effect to be interpreted by some handler later on.
+send :: (Inject effect sig, Carrier sig m) => effect m a -> m a
+send = eff . inj
+{-# INLINE send #-}
--- a/src/Control/Carrier.hs
+++ b/src/Control/Carrier.hs
@@ -17,6 +17,6 @@ import Control.Effect.Sum
 type Has eff sig m = (Inject eff sig, Carrier sig m)
 
 -- | Construct a request for an effect to be interpreted by some handler later on.
-send :: (Inject effect sig, Carrier sig m) => effect m a -> m a
+send :: (Inject eff sig, Carrier sig m) => eff m a -> m a
 send = eff . inj
 {-# INLINE send #-}
--- a/src/Control/Carrier.hs
+++ b/src/Control/Carrier.hs
@@ -4,7 +4,7 @@ module Control.Carrier
   module Control.Carrier.Class
 , module Control.Carrier.Pure
 , module Control.Effect.Class
-, module Control.Effect.Sum
+, (:+:)(..)
 , Has
 , send
 ) where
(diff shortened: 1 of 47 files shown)


src/Control/Carrier.hs
```diff
@@ -1,12 +1,22 @@
...
 import Control.Carrier.Class
 import Control.Carrier.Pure
 import Control.Effect.Class
 import Control.Effect.Sum
+
+type Has eff sig m = (Inject eff sig, Carrier sig m)
+
+-- | Construct a request for an effect to be interpreted by some handler later on.
+send :: Has eff sig m => eff m a -> m a
+send = eff . inj
```
robrix: I moved this back out here because `Has` definitely doesn’t belong in `Control.Effect.Sum`.

src/Control/Carrier/Interpose.hs
```diff
@@ -45,7 +45,7 @@ newtype Handler eff m = Handler (forall x . eff m x -> m x)
 runHandler :: (HFunctor eff, Functor m) => Handler eff m -> eff (ReaderC (Handler eff m) m) a -> m a
 runHandler h@(Handler handler) = handler . hmap (runReader h)
 
-instance (HFunctor eff, Carrier sig m, Member eff sig) => Carrier sig (InterposeC eff m) where
+instance (HFunctor eff, Carrier sig m, Project eff sig) => Carrier sig (InterposeC eff m) where
```
robrix: This is basically the only thing in the library using `Project`.

I want to 🔥 it so badly, but there’s no way @patrickt will let me 😅

robrix: _However_, see #220.

PR #223, opened 2019-09-26
🔥 interposition

### LINKED ISSUE #220: Interposition is redundant to interpretation
`InterposeC` intercepts requests for an effect without extending the signature; instead, it projects out and handles requests for that effect, passing any others along.

`InterpretC` and other typical carriers extend the signature at the left, and handle any requests for their signature, passing any others along.

Both extend the monad stack with another layer.

From the perspective of code run inside the monad, it makes no difference whether the available effects are `A :+: B :+: C` and `B` is interposed, or `B :+: A :+: B :+: C` and requests for `B` are intercepted by the leftmost carrier for `B`. (It is technically possible to observe the difference since you can constrain the signature and pattern match on the effect sum with a carrier, but there’s no point to doing so and so you simply don’t.)

Therefore, interposition as currently implemented is redundant: you could just have a carrier, either using `runInterpret` or manually defined, which extends the signature. Doing so would be quite useful since `InterposeC` is inefficient and complicated, and it would also allow us to remove `prj` altogether.

robrix:
This PR:

- [x] 🔥s the interposition carrier.
- [x] 🔥s the interposition carrier in the tests.
- [x] 🔥s `Project`.
- [x] Renames `Inject` to `Member`.
- [x] Updates the changelog.
- [x] Depends on #219.
- [x] Fixes #220.
- [x] Document the `Member` instances, per https://github.com/fused-effects/fused-effects/pull/219#discussion_r329358954.


src/Control/Effect/Sum.hs
```diff
@@ -20,54 +18,25 @@ instance (HFunctor f, HFunctor g) => HFunctor (f :+: g)
...
 instance {-# OVERLAPPABLE #-}
-         Inject l r
-      => Inject l (l' :+: r) where
+         Member l r
+      => Member l (l' :+: r) where
   inj = R . inj
-
-
-class Project (sub :: (* -> *) -> (* -> *)) sup where
-  prj :: sup m a -> Maybe (sub m a)
```
robrix: `InterposeC` represented the only use of `Project` in the library, so I’ve opted to remove it. The justification for this is a little bit weaker, IMO, but in short: it’s impolite, and probably surprising, to (metaphorically) wander into other people’s houses uninvited.

We currently use `prj` in three places in `semantic` (all of which represent some variation on interposition which should be changed), as well as several places in `semantic-core` and `path` (all of which represent syntactic operations which we could support by redefining `Project`/`prj` locally).

The latter category make me wonder if we should retain this for syntax use cases, but tbh `bound`-style syntax having the same kind as higher-order effects is at best a pun, because the syntax is generally a very different _shape_:

- Nullary constructors do not represent abnormal termination of control.
- Constructors are designed to hold terms, rather than continuations per se.

This makes `Carrier` instances beyond the typical `Term` at best inconvenient, and perhaps warrants a `fused-syntax` package enabling e.g. algebras for non-`Monad`ic carriers. (And perhaps that package and `fused-effects` will someday share some higher-order syntax such as sums. Life is full of possibility.)

Thus, while I’m slightly concerned about this, I am currently, on the balance, convinced that removing it is right for the library; but I would nevertheless particularly welcome a gut check.

test/Control/Effect/Spec.hs
```diff
@@ -62,30 +60,6 @@ instance (Carrier sig m, Effect sig) => Carrier (Reader r :+: sig) (ReinterpretR
   eff (R other)         = ReinterpretReaderC (eff (R (handleCoercible other)))
 
 
-interposition :: Spec
-interposition = describe "interposition" $ do
```
robrix: tbh we could have got rid of this when we added `InterposeC`.

src/Control/Effect/Sum.hs
```diff
@@ -1,14 +1,16 @@
...
+-- | Operations on /sums/, combining effects into a /signature/.
 module Control.Effect.Sum
-( (:+:)(..)
-, Member
-, Inject(..)
-, Project(..)
+( -- * Membership
+  Member(..)
+  -- * Sums
+, (:+:)(..)
```
robrix: The docs read a lot better with `Member` listed first, due to the extremely large number of instances involving `:+:`.

src/Control/Effect/Sum.hs
```diff
@@ -20,54 +22,35 @@ instance (HFunctor f, HFunctor g) => HFunctor (f :+: g)
...
 
-type Member sub sup = (Inject sub sup, Project sub sup)
-
-
-class Inject (sub :: (* -> *) -> (* -> *)) sup where
+-- | The class of types present in a signature.
+--
+--   This is based on Wouter Swierstra’s design described in [Data types à la carte](http://www.cs.ru.nl/~W.Swierstra/Publications/DataTypesALaCarte.pdf). As described therein, overlapping instances are required in order to distinguish e.g. left-occurrence from right-recursion.
+--
+--   It should not generally be necessary for you to define new 'Member' instances, but these are not specifically prohibited if you wish to get creative.
```
robrix: Per https://github.com/fused-effects/fused-effects/pull/219#discussion_r329358954, I’ve documented the overall design, the overlapping instances, and the reassociation; I’ve also noted here that you probably don’t need to define your own instances.

robrix:
> My take on this is that it definitely improves the library, and since we have another answer for the use-case that `Interpose` addresses, it’s a good thing to kill `prj`. However, I don’t think we should merge this until we fix [github/semantic#289](https://github.com/github/semantic/issues/289), because the use of `Interpose` in Semantic is (surprise, surprise) very subtle, and I don’t want to get it wrong when we port semantic to fused-effects-1.0. Does that sound good?

IMO we can just copy in `prj`; `semantic` uses its own definition of interposition and eavesdropping, rather than relying on `fused-effects`’. So I think we should go ahead and merge; copying `prj` in over there is but a `grep` away.

robrix:
We’re gonna go ahead with the merge 👍

PR #236, opened 2019-09-29
Generalize NonDet benchmarks

robrix:
This PR generalizes the `NonDet` benchmarks over the carrier, allowing us to compare performance of multiple implementations.

### DIFF
--- a/benchmark/Bench.hs
+++ b/benchmark/Bench.hs
@@ -12,13 +12,11 @@ import Data.Functor.Identity
 import Data.Monoid (Sum(..))
 import Gauge
 
-import qualified NonDet.NQueens as NQueens
+import qualified NonDet
 
 main :: IO ()
 main = defaultMain
-  [ bgroup "NonDet"
-    [ NQueens.benchmark
-    ]
+  [ NonDet.benchmark
   , bgroup "WriterC"
     [ bgroup "Cod"
       [ bench "100"   $ whnf (run . runCod pure . execWriter @(Sum Int) . runCod pure . tellLoop) 100
--- a/benchmark/NonDet.hs
+++ b/benchmark/NonDet.hs
@@ -1,0 +1,2 @@
+module NonDet
+() where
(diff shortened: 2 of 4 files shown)


benchmark/Bench.hs
```diff
@@ -12,13 +12,11 @@ import Data.Functor.Identity
...
 
-import qualified NonDet.NQueens as NQueens
+import qualified NonDet
 
 main :: IO ()
 main = defaultMain
-  [ bgroup "NonDet"
-    [ NQueens.benchmark
-    ]
+  [ NonDet.benchmark
```
robrix: This group is now provided by `NonDet`.

benchmark/NonDet/NQueens.hs
```diff
@@ -45,12 +45,10 @@ addOne n curr = do
...
-runQueens :: Int -> [Board]
-runQueens = run . runNonDet . queens
-
-benchmark :: Gauge.Benchmark
-benchmark = bgroup "N-queens problem"
-  [ bench "4"  $ whnf runQueens 4
-  , bench "8"  $ whnf runQueens 8
-  , bench "16" $ whnf runQueens 16
+benchmark :: (Alternative m, Monad m) => String -> (m Board -> [Board]) -> Gauge.Benchmark
+benchmark title runQueens = bgroup title
```
robrix: We pass in the title and handler so that we can run the benchmark group against different carriers.

benchmark/NonDet.hs
```diff
@@ -0,0 +1,17 @@
...
+import qualified Control.Carrier.NonDet.Church as NonDet.Church
+import           Control.Carrier.Pure
+import           Gauge hiding (benchmark)
+import qualified NonDet.NQueens as NQueens
+
+benchmark :: Gauge.Benchmark
+benchmark = bgroup "NonDet"
+  [ bgroup "N-queens problem"
+    [ NQueens.benchmark "NonDet.Church" (run . NonDet.Church.runNonDet)
+    , NQueens.benchmark "[]"            (id @[_])
```
robrix: Right now this is what we’ve got; they’re approximately on par.

PR #237, opened 2019-09-29
Tasty

robrix:
This PR migrates tests & examples to use `tasty`, removes some of them, and rearranges some others.

- [x] Migrate the examples to use `tasty`.


examples/Inference.hs
```diff
@@ -0,0 +1,47 @@
+{-# LANGUAGE FlexibleContexts, FlexibleInstances, GeneralizedNewtypeDeriving, MultiParamTypeClasses, TypeApplications, TypeOperators, UndecidableInstances #-}
+module Inference
```
robrix: I brought this over from the unit tests and expanded on it significantly to document the idiom.

test/Control/Effect/Spec.hs
```diff
@@ -1,122 +0,0 @@
...
-spec = do
-  inference
-  reinterpretation
-  interposition
-  fusion
-
-inference :: Spec
-inference = describe "inference" $ do
-  it "can be wrapped for better type inference" $
-    run (runHasEnv (runEnv "i" ((++) <$> askEnv <*> askEnv))) `shouldBe` "ii"
```
robrix: This is a pretty important technique and underserved here, so I moved it into the examples, above.

test/Control/Effect/Spec.hs
```diff
@@ -1,122 +0,0 @@
...
-newtype HasEnv env m a = HasEnv { runHasEnv :: m a }
-  deriving (Applicative, Functor, Monad)
-
-instance Carrier sig m => Carrier sig (HasEnv env m) where
-  eff = HasEnv . eff . handleCoercible
-
-
-reinterpretation :: Spec
-reinterpretation = describe "reinterpretation" $ do
-  it "can reinterpret effects into other effects" $
```
robrix: 1. We’re testing this already in the actual library, e.g. with `FreshC`.

2. This is not a good example; it’s contrived, and ill-justified.

Therefore, I’ve opted to 🔥 it since it’s not serving any purpose here and I’d rather not spend the time rewriting it into an example; this is honestly deserving of a much better-thought-out example (cf #9).

test/Control/Effect/Spec.hs
```diff
@@ -1,122 +0,0 @@
...
-    a <- ReinterpretReaderC get
-    ReinterpretReaderC (put (f a))
-    v <- m
-    ReinterpretReaderC (put a)
-    k v
-  eff (R other)         = ReinterpretReaderC (eff (R (handleCoercible other)))
-
-
-interposition :: Spec
-interposition = describe "interposition" $ do
```
robrix: 1. This exists in the library, so reimplementing it is particularly purposeless.

2. It’s going away in #223 anyhow.

3. It’s also the wrong way to do this.

4. Plus there’s no motivation, it’s a bad example, etc.

🔥

test/Control/Effect/Spec.hs
```diff
@@ -1,122 +0,0 @@
...
-    | Just (Fail s) <- prj op = InterposeC (send (Fail ("hello, " ++ s)))
-    | otherwise               = InterposeC (eff (handleCoercible op))
-
-
-shouldSucceed :: Inspection.Result -> Expectation
-shouldSucceed (Success _) = pure ()
-shouldSucceed (Failure f) = expectationFailure f
-
-fusion :: Spec
-fusion = describe "fusion" $ do
```
robrix: This has been moved into its own module.

test/Control/Effect/NonDet/Spec.hs
```diff
@@ -1,23 +0,0 @@
-module Control.Effect.NonDet.Spec
```
robrix: Moved to `NonDet`.

test/Fusion.hs
```diff
@@ -0,0 +1,51 @@
...
+    @?= Nothing
+  , testCase "eliminates calls to eff" $
+    failureOf $(inspectTest $ 'countDown `doesNotUse` 'eff)
+    @?= Nothing
+  ]
+
+
+failureOf :: Inspection.Result -> Maybe String
+failureOf (Success _) = Nothing
+failureOf (Failure f) = Just f
```
robrix: I really wanted to be able to use `@?=`, so I made this just return the `String`. We’ll see the error message in failures.

test/NonDet.hs
```diff
@@ -0,0 +1,33 @@
...
+
+import Control.Carrier.Error.Either
+import Control.Carrier.NonDet.Church
+import Control.Carrier.State.Strict
+import Prelude hiding (error)
+import Test.Tasty
+import Test.Tasty.HUnit
+
+tests :: TestTree
+tests = testGroup "NonDet"
```
robrix: I really wanted to abstract these over the carrier so that we could test them properly against each carrier independently, but the [ridiculous contortions that required](https://github.com/fused-effects/fused-effects/pull/237/commits/08546e9e13734b13036fa984f4ecdadcc8fbc1de) were not worth the potential benefit.

New plan: can we express interpreting an effect as an effect? 🤔

test/Test.hs
```diff
@@ -0,0 +1,13 @@
+module Main
```
robrix: Renamed this file while I was at it. I don’t like the word “spec” for “test.”

PR #239, opened 2019-09-29
Effect laws

### LINKED ISSUE #142: Effect laws are undocumented

### LINKED ISSUE #249: Where should solo carriers live?
`Reader` only has one obvious carrier (modulo #206), which currently lives in `Control.Carrier.Reader`. This is pretty unsatisfying, tho, because the whole schtick is that we want to admit carriers we haven’t thought of yet, and it feels like it’s unnecessarily confusing for it to be the “default” carrier when e.g. `State` doesn’t have one.

I’d rather move it somewhere else sensible under `Control.Carrier.Reader`, and link to the carrier module from the effect haddocks (cf #244). But what’s a good schema for naming that sort of carrier? Some ideas:

- `Control.Carrier.Reader.Default`
- `Control.Carrier.Reader.Function`
- `Control.Carrier.Reader.Arrow`

Ideally, a schema will adapt well to `Control.Carrier.Resource`, and possibly also `Control.Carrier.Lift` (altho I admit that one feels like a bit of a special case).

robrix:
This PR is aimed at documenting effect laws and verifying them against each defined carrier.

- [x] Fixes #142.
- [x] ~~Fixes #249.~~ Moved the `Reader` carrier back where it was.


fused-effects.cabal
```diff
@@ -117,14 +117,21 @@ test-suite test
   hs-source-dirs:      test
   main-is:             Test.hs
   other-modules:
+    Empty.Maybe
+    Error.Either
     Fusion
-    NonDet
+    NonDet.Church
+    Pure
```
robrix: I’ve started naming test modules after the carriers they test.

src/Control/Carrier/Cull/Church.hs
```diff
@@ -72,6 +72,4 @@ instance (Carrier sig m, Effect sig) => Carrier (Cull :+: NonDet :+: sig) (CullC
 
 
 -- $setup
--- >>> :seti -XFlexibleContexts
--- >>> :seti -XTypeApplications
```
robrix: We now specify these for all the doctests.

src/Control/Carrier/Empty/Maybe.hs
```diff
@@ -27,7 +27,7 @@ runEmpty = runMaybeT . runEmptyC
 {-# INLINE runEmpty #-}
 
 newtype EmptyC m a = EmptyC { runEmptyC :: MaybeT m a }
-  deriving (Applicative, Functor, Monad, MonadFix, MonadIO, MonadTrans)
+  deriving (Applicative, Functor, Monad, MonadFix, MonadIO, MonadTrans, Show)
```
robrix: We now generate these in the tests, and being able to show them is handy (vs. using `Blind` all over the place).

src/Control/Carrier/Interpret.hs
```diff
@@ -113,5 +113,3 @@ instance (HFunctor eff, HFunctor sig, Reifies s (Handler eff m), Monad m, Carrie
 
 -- $setup
 -- >>> import Test.QuickCheck
--- >>> import Control.Effect.Pure
--- >>> import Control.Effect.State
```
robrix: Unused/redundant.

src/Control/Carrier/Pure.hs
```diff
@@ -57,3 +59,7 @@ instance MonadFix PureC where
 instance Carrier Pure PureC where
   eff v = case v of {}
   {-# INLINE eff #-}
+
+
+instance Show1 PureC where
+  liftShowsPrec sp _ d = showsUnaryWith sp "PureC" d . run
```
robrix: Both `MaybeT m` and `ExceptT e m` require `Show1 m` for their `Show` instances.

src/Control/Effect/Empty.hs
```diff
@@ -22,7 +24,11 @@ instance Effect   Empty
 
 -- | Abort the computation.
 --
---   prop> run (runEmpty empty) === Nothing
+-- 'empty' annihilates '>>=':
+--
+-- @
+-- 'empty' '>>=' k = 'empty'
+-- @
```
robrix: We now document the general law instead of running a specific property test.

src/Control/Effect/Empty.hs
```diff
@@ -32,7 +38,8 @@ guard True  = pure ()
...
 
--- $setup
--- >>> :seti -XFlexibleContexts
--- >>> import Test.QuickCheck
--- >>> import Control.Carrier.Empty.Maybe
+-- Properties
+
+-- | 'empty' annihilates '>>='.
+empty_annihilation :: Has Empty sig m => (m b -> m b -> prop) -> (a -> m b) -> prop
+empty_annihilation (===) k = (empty >>= k) === empty
```
robrix: We additionally define and export properties parameterized by the equality test. This property can be used with QuickCheck, hedgehog, and presumably other property testing frameworks, and doesn’t require the library to depend on any of them.

The intention is for every `Empty` carrier to test its conformance to this law.

test/Doctest.hs
```diff
@@ -6,6 +6,11 @@ import System.Environment
...
-main = do
-  args <- getArgs
-  doctest ("-isrc" : "--fast" : if null args then ["src"] else args)
+main = getArgs >>= doctest . build where
+  build args
+    = "-isrc"
+    : "--fast"
+    : "-XFlexibleContexts"
+    : "-XFlexibleInstances"
+    : "-XTypeApplications"
```
robrix: These are now available to every module’s doctests.

test/Empty/Maybe.hs
```diff
@@ -0,0 +1,30 @@
...
+    \ k -> empty_annihilation ((~=) @B) (applyFun @A k)
+  ]
+
+(~=) :: (Eq a, Show a) => EmptyC PureC a -> EmptyC PureC a -> Property
+m1 ~= m2 = run (runEmpty m1) === run (runEmpty m2)
+
+
+instance Arbitrary1 m => Arbitrary1 (EmptyC m) where
+  liftArbitrary genA = EmptyC . MaybeT <$> liftArbitrary @m (liftArbitrary @Maybe genA)
+  liftShrink shrinkA = map (EmptyC . MaybeT) . liftShrink (liftShrink shrinkA) . runEmpty
```
robrix: This is the direct approach. Another approach that occurred to me would be building up programs using the constructors of the effects for the various carriers (in this case `empty` and `pure`) and binding/sequencing chains of them together. That would certainly work, but ultimately every such program ends up being represented as `m (Maybe a)`, so we can, and do, generate that directly and wrap it up instead.

test/Error/Either.hs
```diff
@@ -0,0 +1,32 @@
+{-# LANGUAGE ScopedTypeVariables, TypeApplications #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+module Error.Either
+( tests
+) where
+
+import Control.Carrier.Error.Either
+import Control.Monad.Trans.Except
+import Pure
+import Test.QuickCheck.Poly
```
robrix: This gives us convenient `A`, `B`, and `C` types with `Arbitrary` instances already attached.

src/Control/Effect/Reader.hs
```diff
@@ -39,30 +41,39 @@ instance Effect (Reader r) where
...
 
--- $setup
--- >>> import Test.QuickCheck
--- >>> import Control.Carrier.Reader
+-- Properties
+
+-- | 'ask' returns the environment variable.
+--
+-- @since 1.0.0.0
+ask_environment :: Has (Reader r) sig m => (b -> b -> prop) -> (r -> m a -> b) -> r -> (r -> m a) -> prop
```
mbj: Just curious, why you are snake case over camel case here? There must be a convention I miss.

robrix: It’s the convention used for e.g. QuickCheck properties to make them easily discoverable using Template Haskell.

test/Gen.hs
```diff
@@ -0,0 +1,176 @@
...
+genT :: Gen (T a)
+genT = T <$> integral (linear 0 100)
+
+newtype T a = T { unT :: Integer }
+  deriving (Enum, Eq, Generic, Num, Ord, Real, Vary)
+
+instance Arg (T a)
+
+instance KnownSymbol s => Show (T s) where
+  showsPrec d = showsUnaryWith showsPrec (symbolVal (Proxy @s)) d . unT
```
robrix: `Test.QuickCheck.Poly` offered `A`, `B`, and `C` newtypes for convenient testing of polymorphic properties. I figured I could improve on them a bit.

1. `Arg` and `Vary` are used by `hedgehog-fn` to generate functions.

2. `T` has a phantom, poly-kinded type parameter enabling us to reuse a single type for multiple different “roles”—e.g. if a property requires handlers, continuations, variables, and actions, the generators for all of these can involve the single polymorphic generator for `T` at some level.

3. The `Show` instance for `T` uses the type parameter for the constructor name, enabling us to show a value of `T { unT = 0 } :: T "A"` much more succinctly _and_ readably as `A 0`. This makes it obvious at a glance which role a value in a test failure was filling, without having to sort out which `forAll` in the failure message corresponds to which variable in the property.

test/Gen.hs
```diff
@@ -0,0 +1,176 @@
...
+
+a :: Gen A
+a = genT
+
+type A = T "A"
+
+b :: Gen B
+b = genT
+
+type B = T "B"
```
robrix: We provide types and generators for computations’ return types…

test/Gen.hs
```diff
@@ -0,0 +1,176 @@
...
+
+e :: Gen E
+e = genT
+
+type E = T "E"
```
robrix: …errors…

test/Gen.hs
```diff
@@ -0,0 +1,176 @@
...
+
+s :: Gen S
+s = genT
+
+type S = T "S"
+
+w :: Gen W
+w = list (linear 0 100) genT
+
+type W = [T "W"]
```
robrix: …and reader, state, and writer variables.

test/Gen.hs
```diff
@@ -0,0 +1,176 @@
...
+
+data Rec as where
+  Nil :: Rec '[]
+  (:.) :: a -> Rec as -> Rec (a ': as)
+
+forall :: (Forall g f, HasCallStack) => g -> f -> Hedgehog.Property
+forall g f = withFrozenCallStack $ Hedgehog.property (forall' g f)
+
+class Forall g f | g -> f, f -> g where
+  forall' :: HasCallStack => g -> f -> PropertyT IO ()
```
robrix: I missed the convenience QuickCheck offered of specifying a property as a function from the generated values to the (in)equality to be tested. `forall` takes a heterogeneous list (`Rec`) of generators, generates a value from each in turn, and calls its second argument with these values. This is far more succinct than generating values using `do`-notation, while still allowing the property to label and classify values.

test/Gen.hs
```diff
@@ -0,0 +1,176 @@
...
+showingFn = With' . flip showsPrec <*> apply
+
+atom :: String -> a -> With a
+atom s = With' (\ _ -> showString s)
+
+liftWith :: String -> (a -> b) -> With a -> With b
+liftWith s w a = atom s w <*> a
+
+liftWith2 :: String -> (a -> b -> c) -> With a -> With b -> With c
+liftWith2 s w a b = atom s w <*> a <*> b
```
robrix: The `With` DSL—named for `forAllWith`, which it subsumes—enables us to build descriptions compositionally, just like `Gen` allows us to build shrinks compositionally. (I looked into `MonadGen` to see if I could generate these by construction there; I cannot.)

- `atom` describes nullary operators.
- `liftWith` & `liftWith2` describe _and apply_ unary & binary operators, respectively.
- `showing` describes primitive values using a `Show` instance.
- `showingFn` describes functions generated using `hedgehog-fn` using `Show` instances.
- the `Applicative` instance combines the descriptions by juxtaposition, yielding descriptions of Haskell function application with appropriate precedence. At present, prefix notation is used for all applications, including infix operators (which should be written with parentheses).

As a result, if we generate a computation `local f m`, we’ll see `local f m` in the output—with one caveat that I’m aware of: the `Show` instance `hedgehog-fn` provides for `Fn` will show output such as `A 0 -> empty _ -> pure (B 1)` for a function which operates as `\case { A 0 -> empty ; _ -> pure (B 1) }`, and to the best of my knowledge it will not parenthesize it appropriately. I may submit a PR to resolve this at some point, or simply wrap `Fn` for our own purposes.

Without this, the vast majority of generated computations can’t be shown as anything more interesting than `_`, which makes debugging and resolving test failures _incredibly_ difficult. This difficulty will be significantly compounded by the extension of `m`, above, to generate monadically-sequenced chains of operations.

test/Gen.hs
```diff
@@ -0,0 +1,176 @@
...
+
+pattern Fn :: (a -> b) -> Fn a b
+pattern Fn f <- (apply -> f)
+
+{-# COMPLETE Fn #-}
+
+pattern FnWith :: (a -> b) -> Fn a (With b)
+pattern FnWith f <- (fmap getWith . apply -> f)
+
+{-# COMPLETE FnWith #-}
```
robrix: Working with `With` and `Fn` required a bunch of unwrapping which was making the properties harder to read, so I defined these pattern synonyms to allow me to unwrap the values once, on the left-hand side of the lambda, and then use the values freely inside, while still getting the benefit of the `Show` instances.

I had initially thought about using a typeclass with an associated type family alongside `Forall` to unwrap e.g. a generated `Fn a b` to `a -> b` before passing it to the property, but this turned out to be quite a convenient, and somewhat less magical, alternative.

test/NonDet.hs
```diff
@@ -1,33 +1,48 @@
...
-    @?= [('a', 'z'), ('b', 'b'), ('a', 'a')]
-  , testCase "collapses results of effects run outside it" $
-    run (runState 'a' (runNonDetA state))
-    @?= ('b', "zbb")
-  , testCase "collects results from higher-order effects run inside it" $
-    run (runNonDetA (runError error))
-    @?= [Right 'z', Right 'a' :: Either Char Char]
-  , testCase "collapses results of higher-order effects run outside it" $
-    run (runError (runNonDetA error))
-    @?= (Right "a" :: Either Char String)
```
robrix: These now live in `NonDet.Church`.

test/Writer.hs
```diff
@@ -0,0 +1,63 @@
...
+  .  (Has (Writer w) sig m, Arg w, Show a, Show w, Vary w)
+  => Gen w
+  -> (forall a . Show a => Gen a -> Gen (With (m a)))
+  -> Gen a
+  -> Gen (With (m a))
+gen w m a = choice
+  [ liftWith2 "(<$)" (<$) . showing <$> a <*> (liftWith "tell" tell . showing <$> w)
+  , subtermM (m a) (\ m -> choice
+    [(\ f -> (liftWith2 "fmap" fmap (liftWith2 "(.)" (.) (showingFn f) (atom "fst" fst)) (liftWith "listen" (listen @w) m))) <$> fn a
+    , pure (liftWith2 "fmap" fmap (atom "snd" snd) (liftWith "listen" (listen @w) m))
```
robrix: `listen` is an interesting case since it produces pairs `(w, a)`. We could conceivably generate functions `(w, a) -> a`, but I opted to generate `w -> a` and invoke `fst`/`snd` instead. I’m not sure it makes much difference.

src/Control/Carrier/Class.hs
```diff
@@ -75,6 +76,9 @@ instance (Carrier sig m, Effect sig) => Carrier (Error e :+: sig) (Except.Except
   eff (L (R (Catch m h k))) = Except.catchE m h >>= k
   eff (R other)             = Except.ExceptT $ eff (handle (Right ()) (either (pure . Left) Except.runExceptT) other)
 
+instance Carrier sig m => Carrier sig (Identity.IdentityT m) where
+  eff = Identity.IdentityT . eff . handleCoercible
+
```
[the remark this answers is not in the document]
patrickt: Should we document this instance somewhere?

src/Control/Carrier/State/Lazy.hs
```diff
@@ -34,7 +31,7 @@ instance Functor m => Functor (StateC s m) where
   fmap f m = StateC $ \ s -> fmap (\ ~(s', a) -> (s', f a)) $ runStateC m s
   {-# INLINE fmap #-}
 
-instance (Functor m, Monad m) => Applicative (StateC s m) where
```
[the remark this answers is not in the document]
patrickt: GHC bug?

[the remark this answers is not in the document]
robrix: Perhaps!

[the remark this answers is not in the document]
robrix: Perhaps; not sure.

src/Control/Carrier/Class.hs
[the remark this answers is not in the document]
patrickt: Fair enough. I don’t think I’ve ever seen `IdentityT` before.

src/Control/Effect/Choose.hs
```diff
@@ -49,24 +64,43 @@ infixl 3 <|>
...
+-- @
+--
 -- @since 1.0.0.0
 many :: Has Choose sig m => m a -> m [a]
 many a = go where go = (:) <$> a <*> go <|> pure []
 
 -- | One or more.
 --
+-- @
+-- 'some' m = (:) '<$>' m '<*>' 'many' m
```
patrickt: My favorite duality!

src/Control/Effect/Fresh.hs
```diff
@@ -35,24 +37,20 @@ instance Effect Fresh where
 
 -- | Produce a fresh (i.e. unique) 'Int'.
 --
---   prop> run (runFresh (replicateM n fresh)) === nub (run (runFresh (replicateM n fresh)))
+-- @
+-- m '>>' 'fresh' ≠ m '>>' 'fresh' '>>' 'fresh'
+-- @
```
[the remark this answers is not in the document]
patrickt: Fair enough. You could do `==` and `succ` but this is fine.

test/Catch.hs
```diff
@@ -0,0 +1,36 @@
...
+gen
+  :: forall e m a sig
+  .  (Has (Catch e) sig m, Arg e, Show a, Show e, Vary e)
+  => Gen e
+  -> (forall a . Show a => Gen a -> Gen (With (m a)))
+  -> Gen a
+  -> Gen (With (m a))
+gen _ m a = do
+  h <- fn @e (m a)
+  subterm (m a) $ \ m -> liftWith2 "catchError" catchError m (fmap getWith <$> showingFn h)
```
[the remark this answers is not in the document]
patrickt: This is so cool!

test/Error.hs
```diff
@@ -0,0 +1,46 @@
...
+  :: (Has (Error e) sig m, Arg a, Arg e, Eq a, Eq b, Eq e, Show a, Show b, Show e, Vary a, Vary e)
+  => Gen e
+  -> (forall a . Show a => Gen a -> Gen (With (m a)))
+  -> Gen a
+  -> Gen b
+  -> (forall a . m a -> PureC (Either e a))
+  -> [TestTree]
+test e m a b runError
+  =  Throw.test e m a b runError
+  ++ Catch.test e m a b runError
```
patrickt: Nitpick: I prefer `<>`.

robrix: I don’t, in this case:

1. it’s unnecessary polymorphism
2. it requires another import in ghc 8.2.2.

src/Control/Carrier/Class.hs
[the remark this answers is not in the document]
robrix: @patrickt: all it means is that `IdentityT` can occur in a carrier stack so long as the monad under it is a carrier.

src/Control/Effect/Fresh.hs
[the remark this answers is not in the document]
robrix: In fact no: the _only_ constraint I wish to place on `fresh` is that it doesn’t produce the same value twice—not the ordering, not that it’s monotone, nothing like that. (I actually kind of regret specializing it to `Int`.)

test/Catch.hs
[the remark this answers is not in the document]
robrix: Thank you!

src/Control/Carrier/State/Lazy.hs
```diff
@@ -95,7 +99,9 @@ runState s c = runStateC c s
 
 -- | Run a lazy 'State' effect, yielding the result value and discarding the final state.
 --
---   prop> run (evalState a (pure b)) === b
+-- @
+-- 'evalState' s m = 'fmap' 'snd' ('runState' s m)
```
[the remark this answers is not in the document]
robrix: s/definitions/specifications 😉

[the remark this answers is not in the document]
robrix: (tho yes they’re absolutely correct by definition as a result of that coincidence)

PR #247, opened 2019-10-04
Split Error into Throw & Catch effects

### LINKED ISSUE #154: Polymorphic `Fail` effect?

robrix:
This PR explores splitting `Error` up into separate `Throw` and `Catch` effects.

- [x] Redefines `Error e` as a synonym for `Throw e :+: Catch e`.
- [x] Redefines `Fail` as a synonym for `Throw String`.
- [x] Fixes #154.

### DIFF
--- a/src/Control/Carrier.hs
+++ b/src/Control/Carrier.hs
@@ -15,9 +15,11 @@ import Control.Effect.Class
 import Control.Effect.Sum
 import Data.Kind (Constraint)
 
-type family Has sub sup m :: Constraint where
-  Has (l :+: r) u m = (Has l u m, Has r u m)
-  Has t         u m = (Member t u, Carrier u m)
+type Has eff sig m = (Has' eff sig, Carrier sig m)
+
+type family Has' sub sup :: Constraint where
+  Has' (l :+: r) u = (Has' l u, Has' r u)
+  Has' t         u = Member t u
 
 
 -- | Construct a request for an effect to be interpreted by some handler later on.
--- a/src/Control/Carrier.hs
+++ b/src/Control/Carrier.hs
@@ -15,6 +15,7 @@ import Control.Effect.Class
 import Control.Effect.Sum
 import Data.Kind (Constraint)
 
+-- | The @m@ is a carrier for @sig@ containing @eff@.
 type Has eff sig m = (Has' eff sig, Carrier sig m)
 
 type family Has' sub sup :: Constraint where
--- a/src/Control/Carrier.hs
+++ b/src/Control/Carrier.hs
@@ -17,12 +17,12 @@ import Control.Effect.Sum
 import Data.Kind (Constraint)
 
 -- | The @m@ is a carrier for @sig@ containing @eff@.
-type Has eff sig m = (Has' eff sig, Carrier sig m)
+type Has eff sig m = (Members eff sig, Carrier sig m)
 
 -- | Decompose sums on the left into multiple 'Member' constraints.
-type family Has' sub sup :: Constraint where
-  Has' (l :+: r) u = (Has' l u, Has' r u)
-  Has' t         u = Member t u
+type family Members sub sup :: Constraint where
+  Members (l :+: r) u = (Members l u, Members r u)
+  Members t         u = Member t u
 
 
 -- | Construct a request for an effect to be interpreted by some handler later on.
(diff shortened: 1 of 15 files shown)


src/Control/Carrier.hs
```diff
@@ -1,22 +1,31 @@
...
 
-type Has eff sig m = (Member eff sig, Carrier sig m)
+-- | @m@ is a carrier for @sig@ containing @eff@.
+--
+-- Note that if @eff@ is a sum, it will be decomposed into multiple 'Member' constraints. While this technically allows one to combine multiple unrelated effects into a single 'Has' constraint, doing so has two significant drawbacks:
+--
+-- 1. Due to [a problem with recursive type families](https://gitlab.haskell.org/ghc/ghc/issues/8095), this can lead to significantly slower compiles.
+--
+-- 2. It defeats @ghc@’s warnings for redundant constraints, and thus can lead to a proliferation of redundant constraints as code is changed.
+type Has eff sig m = (Members eff sig, Carrier sig m)
```
robrix: We’ve had a rather nasty gotcha lurking since #219 that wasn’t discovered until now: Using a `Has` constraint with an effect defined as a sum wouldn’t allow `ghc` to solve `Member` constraints for the members of that sum. Basically, given `Member (A :+: B) sig`, `ghc` wasn’t able to deduce `Member A sig`, which is necessary for `Has A sig m` to be solved.

This remained latent because `NonDet`, the only other effect defined as a sum of other effects, is used exclusively via the `Alternative` interface—we never use `Has NonDet sig m` constraints. It surfaced now because the common pattern is to use `Error e` even when you only need `Throw e`—in no small part because it wasn’t possible to say solely `Throw e` until this PR.

The solution is to decompose the sum into a series of individual `Member` constraints. Note that this technically makes `Has` capable of expressing multiple unrelated effects in a single constraint (synonym):

```haskell
foo :: Has (Reader r :+: State i :+: Error e :+: Writer w) sig m => m ()
```

However, as documented here and on `Members`, this can cause performance problems as well as defeating warnings, so it’s best to avoid it.

src/Control/Effect/Catch.hs
```diff
@@ -0,0 +1,30 @@
...
+  handle state handler (Catch m h k) = Catch (handler (m <$ state)) (handler . (<$ state) . h) (handler . fmap k)
+
+
+-- | Run a computation which can throw errors with a handler to run on error.
+--
+-- Errors thrown by the handler will escape up to the nearest enclosing 'catchError' (if any).
+-- Note that this effect does /not/ handle errors thrown from impure contexts such as IO,
+-- nor will it handle exceptions thrown from pure code. If you need to handle IO-based errors,
+-- consider if 'Control.Effect.Resource' fits your use case; if not, use 'liftIO' with
+-- 'Control.Exception.try' or use 'Control.Exception.Catch' from outside the effect invocation.
```
robrix: I’ve removed the property tests for the time being; they’re being moved in #239 anyway.

src/Control/Effect/Fail.hs
```diff
@@ -1,18 +1,12 @@
...
+import Control.Effect.Throw
 import qualified Control.Monad.Fail as Fail
-import GHC.Generics (Generic1)
 
-newtype Fail (m :: * -> *) k = Fail String
-  deriving (Functor, Generic1)
-
-instance HFunctor Fail
-instance Effect   Fail
+type Fail = Throw String
```
robrix: There is no longer any reason for `Fail` to be distinct from `Throw`!

src/Control/Carrier/Throw/Either.hs
```diff
@@ -0,0 +1,32 @@
...
+import Control.Monad.Fix
+import Control.Monad.IO.Class
+import Control.Monad.Trans.Class
+
+-- | Run a 'Throw' effect, returning failures in 'Left' and successful computations’ results in 'Right'.
+runThrow :: ThrowC e m a -> m (Either e a)
+runThrow = runError . runThrowC
+
+newtype ThrowC e m a = ThrowC { runThrowC :: ErrorC e m a }
+  deriving (Alternative, Applicative, Functor, Monad, Fail.MonadFail, MonadFix, MonadIO, MonadPlus, MonadTrans)
```
robrix: You can, of course, also just use `ErrorC` directly, but maybe you want to guarantee the absence of `Catch` from the signature or something.

robrix:
I am really quite happy with how this one turned out ✨

PR #266, opened 2019-10-15
Classify generated computations

robrix:
This PR labels the computations generated during testing, in order to illustrate that we’re getting good coverage. It also makes some other simplifications to the testing, generating, &c. APIs.


test/Catch.hs
```diff
@@ -11,26 +11,24 @@ import Test.Tasty.Hedgehog
...
+  .  (Has (Catch e) sig m, Arg e, Show e, Vary e)
   => Gen e
-  -> (forall a . Show a => Gen a -> Gen (With (m a)))
+  -> (forall a . Gen a -> Gen (m a))
   -> Gen a
-  -> Gen (With (m a))
-gen _ m a = do
-  h <- fn @e (m a)
-  subterm (m a) $ \ m -> liftWith2 "catchError" catchError m (fmap getWith <$> showingFn h)
+  -> Gen (m a)
```
robrix: 1. We no longer require `Show` constraints.

2. We no longer require `With`.

Both are due to the fact that `Gen` is now a monad we construct wrapping `Hedgehog.Gen (Term a)` (`Term` being a deep embedding of the prior `With` logic).

test/Catch.hs
```diff
@@ -11,26 +11,24 @@ import Test.Tasty.Hedgehog
...
+gen _ m a = label "catchError" catchError <*> m a <*> fn @e (m a)
```
robrix: We now have much simpler combinators for generating showable-by-construction terms:

- `atom` & `label` construct prefix terms
- `infixL` & `infixR` construct infix terms
- `<*>` does function application (as ever)

test/Catch.hs
```diff
@@ -11,26 +11,24 @@ import Test.Tasty.Hedgehog
...
-  -> (forall a . Show a => Gen a -> Gen (With (m a)))
+  -> (forall a . Gen a -> Gen (m a))
   -> Gen a
   -> Gen b
   -> (forall a . m a -> PureC (Either e a))
   -> [TestTree]
 test e m a _ runCatch =
   [ testProperty "catchError intercepts throwError" . forall (e :. fn (m a) :. Nil) $
-    \ e (FnWith h) -> runCatch (throwError e `catchError` h) === runCatch (h e)
+    \ e h -> runCatch (throwError e `catchError` h) === runCatch (h e)
```
robrix: `fn` is now specialized to our new definition of `Gen`, and produces a (showable) function directly, rather than requiring any sort of unwrapping.

test/Choose.hs
```diff
@@ -20,23 +20,23 @@ tests = testGroup "Choose"
...
 gen
-  :: (Has Choose sig m, Show a)
-  => (forall a . Show a => Gen a -> Gen (With (m a)))
+  :: Has Choose sig m
+  => (forall a . Gen a -> Gen (m a))
   -> Gen a
-  -> Gen (With (m a))
-gen m a = subterm2 (m a) (m a) (liftWith2 "(<|>)" (<|>))
+  -> Gen (m a)
+gen m a = addLabel "<|>" (infixL 3 "<|>" (<|>) <*> m a <*> m a)
```
robrix: There’s no automatically labelling infix constructor—there are only two places we want to label infix operators—so we use `addLabel` to automatically classify the term.

test/Gen.hs
```diff
@@ -21,60 +21,72 @@ module Gen
...
+  -> Gen (m a)                                                          -- ^ A generator producing computations.
 m with = go where
-  go :: forall a . Show a => Gen a -> Gen (With (m a))
-  go a = recursive choice [(liftWith "pure" pure . showing) <$> a] [with go a]
+  go :: forall a . Gen a -> Gen (m a)
+  go a = Gen $ recursive Hedgehog.Gen.choice
+    [ runGen (Gen.label "pure" pure <*> a) ]
+    [ frequency
+      [ (3, runGen (with go a))
+      , (1, runGen (addLabel ">>" (infixL 1 ">>" (>>) <*> go a <*> go a)))
```
robrix: We construct monadic sequences of terms now.

test/Gen.hs
```diff
@@ -21,60 +21,72 @@ module Gen
...
 
-instance Arg (T a)
+instance Fn.Arg (T a)
+
+instance S.Semigroup (T a) where
+  T a <> T b = T (a + b)
+
+instance Monoid (T a) where
+  mempty = T 0
+  mappend = (S.<>)
```
robrix: Defining these instances allows us to avoid using `[]` for `w`, which means much more efficient writer tests.

test/Gen.hs
```diff
@@ -105,9 +117,15 @@ s = genT
 type S = T "S"
 
 w :: Gen W
-w = list (linear 0 100) genT
+w = genT
+
+type W = T "W"
+
+fn :: (Fn.Arg a, Fn.Vary a, Show a) => Gen b -> Gen (a -> b)
+fn b = Gen (lift (fmap (fmap runTerm) . showingFn <$> Fn.fn (fst <$> runWriterT (runGen b))))
```
robrix: This definition regrettably discards labels, because there’s no way to plumb them through `Fn.fn` or its limitation to `Hedgehog.Gen`.

test/Gen.hs
```diff
@@ -105,9 +117,15 @@ s = genT
...
 
-type W = [T "W"]
+choice :: [Gen a] -> Gen a
+choice = Gen . Hedgehog.Gen.choice . Prelude.map runGen
```
robrix: We don’t use many of hedgehog’s combinators in the property tests, but we do use `choice`, so I’ve wrapped it up.

test/Gen.hs
```diff
@@ -125,52 +143,72 @@ class Forall g f | g -> f, f -> g where
 instance Forall (Rec '[]) (PropertyT IO ()) where
   forall' Nil = id
 
-instance (Forall (Rec gs) b, Show a) => Forall (Rec (Gen a ': gs)) (a -> b) where
+instance (Forall (Rec gs) b) => Forall (Rec (Gen a ': gs)) (a -> b) where
   forall' (g :. gs) f = do
-    a <- Hedgehog.forAll g
-    forall' gs (f a)
+    (a, labels) <- Hedgehog.forAll (runWriterT (runGen g))
+    traverse_ Hedgehog.label labels
```
robrix: We now produce the labels for each generator and classify the property with them.

test/Gen.hs
```diff
@@ -125,52 +143,72 @@ class Forall g f | g -> f, f -> g where
...
-atom :: String -> a -> With a
-atom s = With' (\ _ -> showString s)
 
-liftWith :: String -> (a -> b) -> With a -> With b
-liftWith s w a = atom s w <*> a
+data Term a where
+  Pure :: (Int -> ShowS) -> a -> Term a
+  InfixL :: Int -> String -> (a -> b -> c) -> Term (a -> b -> c)
+  InfixR :: Int -> String -> (a -> b -> c) -> Term (a -> b -> c)
+  (:<*>) :: Term (a -> b) -> Term a -> Term b
```
robrix: I decided this all wasn’t over-engineered enough, so I rewrote `With` as a deep embedding, et voilà, `Term`.

The main takeaway is that this allows us to do smarter partial application—basically, operator sections with correct precedence. It also makes it easier and more concise to combine labels, resulting in much briefer, clearer code. None of those `liftWith*`s cluttering things up!

test/Gen.hs
```diff
@@ -125,52 +143,72 @@ class Forall g f | g -> f, f -> g where
...
+  f :<*> a -> runTerm f $ runTerm a
 
-pattern With :: a -> With a
-pattern With a <- (getWith -> a)
+instance Functor Term where
+  fmap = liftA
 
-{-# COMPLETE With #-}
+instance Applicative Term where
+  pure = Pure (const (showString "_"))
```
robrix: `pure` (and hence also `fmap`) can’t show the value. However…

test/Gen.hs
```diff
@@ -125,52 +143,72 @@ class Forall g f | g -> f, f -> g where
...
+instance Show (Term a) where
+  showsPrec d = \case
+    Pure s _ -> s d
+    InfixL _ s _ -> showParen True (showString s)
+    InfixR _ s _ -> showParen True (showString s)
+    InfixL p s _ :<*> a :<*> b -> showParen (d > p) (showsPrec p a . showString " " . showString s . showString " " . showsPrec (succ p) b)
+    InfixR p s _ :<*> a :<*> b -> showParen (d > p) (showsPrec (succ p) a . showString " " . showString s . showString " " . showsPrec p b)
+    InfixL p s _ :<*> a -> showParen True (showsPrec p a . showString " " . showString s)
+    InfixR p s _ :<*> a -> showParen True (showsPrec (succ p) a . showString " " . showString s)
+    f :<*> a -> showParen (d > 10) (showsPrec 10 f . showString " " . showsPrec 11 a)
```
robrix: …everything else can be shown correctly:

- `Pure` embeds a precedence-accurate `Int -> ShowS` to show atoms
- `InfixL`/`InfixR` encode precedence for an infix operation
- applications of infix operators to two operands nest correctly
- applications of infix operators to one operand are properly parenthesized as operator sections
- bare infix operators are parenthesized as atoms
- prefix applications are left-associated and nest correctly

This even extends to uses of infix operations on the left of triple applications, e.g. `(f <$> a) b`.

It doesn’t know about algebraic simplifications, of course—it’s not going to apply the fmap fusion law for you—but it’s ridiculously effective for what it is, and also embeds the values so they can be extracted, combined, and used.

test/Gen.hs
```diff
@@ -125,52 +143,72 @@ class Forall g f | g -> f, f -> g where
...
 
-{-# COMPLETE Fn #-}
 
-pattern FnWith :: (a -> b) -> Fn a (With b)
-pattern FnWith f <- (fmap getWith . apply -> f)
+newtype Gen a = Gen { runGen :: WriterT (Set.Set LabelName) Hedgehog.Gen (Term a) }
+  deriving (Functor)
```
robrix: We use `WriterT` for this because `WriterT` has a `MonadGen` instance.

test/Gen.hs
```diff
@@ -125,52 +143,72 @@ class Forall g f | g -> f, f -> g where
...
 
-{-# COMPLETE FnWith #-}
+instance Applicative Gen where
+  pure = Gen . pure . pure
+  Gen m1 <*> Gen m2 = Gen ((<*>) <$> m1 <*> m2)
```
robrix: Note that we don’t offer a `Monad` instance for `Gen`—I haven’t figured out whether we can have good support for `Term`, and if so, how.

PR #270, opened 2019-10-16
Experiment: suspend carrier

robrix:
This effect-polymorphic carrier is one potential alternative to using `Resumable`.

### DIFF
--- a/fused-effects.cabal
+++ b/fused-effects.cabal
@@ -65,6 +65,7 @@ library
     Control.Carrier.Resumable.Resume
     Control.Carrier.State.Lazy
     Control.Carrier.State.Strict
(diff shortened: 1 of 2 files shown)


ocharles:
Can you really do much with a suspended effect this way? You end existentially quantifying what the effect returns, so I guess you can resume it, but can't really see what it finally returns?

robrix:
@ocharles: Not much, but about as much as the error types we use with `Resumable` in `semantic`: log failures in the tests, say.

I wouldn’t consider this either a) complete or b) indicative of much thought 😊 really just a quick spike at one way to replace `Resumable`.

robrix:
This clearly works, but I’m not keen on having this in the library.

PR #273, opened 2019-10-19
Monad properties

### LINKED ISSUE #229: Laws are untested
Carriers’ adherence to the `Applicative`, `Monad`, &c. laws is largely untested, leading to bugs like `ErrorC` having an unlawful `Applicative` instance (cf #230).

We should test them.

robrix:
This PR:

- [x] Defines properties for the monad laws.
- [x] Tests them against all the carriers.
- [x] Fixes #229.


test/Choose.hs
```diff
@@ -7,16 +7,22 @@ module Choose
...
-  [ testGroup "ChooseC"  $ test (m gen) a b (ChooseC.runChooseS (pure . pure))
-  , testGroup "NonEmpty" $ test (m gen) a b (pure . toList)
-  ]
+  [ test "ChooseC"  (ChooseC.runChooseS (pure . pure))
+  , test "NonEmpty" (pure . toList)
+  ] where
+  test :: Has Choose sig m => String -> (forall a . m a -> PureC [a]) -> TestTree
+  test name run = testGroup name
+    $  Monad.test  (m gen) a b c (pure (Identity ())) (run . runIdentity)
+    ++ Choose.test (m gen) a b                        run
```
robrix: We test the monad properties here rather than adding them to the `test` function below so that composed tests (like `NonDet`) won’t run the monad tests multiple times. It also avoids the need to pass extra parameters to the effect tests to accommodate the monad tests.

test/Monad.hs
```diff
@@ -0,0 +1,31 @@
...
+import Test.Tasty.Hedgehog
+
+test
+  :: (Monad m, Arg a, Arg b, Eq (f a), Eq (f b), Eq (f c), Show a, Show b, Show c, Show (f a), Show (f b), Show (f c), Vary a, Vary b, Functor s)
+  => (forall a . Show a => Gen a -> Gen (m a))
+  -> Gen a
+  -> Gen b
+  -> Gen c
+  -> Gen (s ())
+  -> (forall a . s (m a) -> PureC (f a))
```
robrix: Passing in a generator for an input functor `s` here allows us to test initial state for carriers like `StateC` and `FreshC`. This accommodates arbitrary functors, but pairs are the only ones we use right now.

Taking a generator for the unit state allows us to avoid having to pass in a constructor function building input state from arbitrary parameters; this is the same trick as is used by `Effect` to pass the initial state separately from the effect.

Issue #279, opened 2019-10-20
Cut.Church: Higher-order effects allow backtracking past cutfail


robrix:
`cutfail` is supposed to prune backtracking:

```haskell
λ run (runReader () (runCutA (cutfail <|> pure 'a'))) :: String
"" 
```

but the `Carrier` instance for `CutC` doesn’t preserve that property when running the effect inside higher-order positions:

```haskell
λ run (runReader () (runCutA (local (id @()) cutfail <|> pure 'a'))) :: String
"a" 
```

PR #283, opened 2019-10-21
Test laws compositionally

robrix:
This PR does a bunch of legwork to test the various laws compositionally, i.e. in relation to other effects & carriers, and spikes this out in particular for `CutC`, as part of investigation into #279.

I’ve also made some effort to split up the generators for each effect’s actions into recursive & nonrecursive portions so we get better coverage of first-order operations.

### DIFF
--- a/test/Cut.hs
+++ b/test/Cut.hs
@@ -26,22 +26,14 @@ tests = testGroup "Cut"
     , testCut
     ] >>= ($ runL CutC.runCutA)
   , testGroup "ReaderC · CutC" $
-    [ Cut.test
-    , testCutfailLocal
-    ] >>= \ f -> f (m genCutReader) a b (pair <*> r <*> unit) (Run (CutC.runCutA . uncurry runReader))
+    Cut.test (Hom (local (id @R))) (m genCutReader) a b (pair <*> r <*> unit) (Run (CutC.runCutA . uncurry runReader))
   , testGroup "CutC · ReaderC" $
-    [ Cut.test
-    , testCutfailLocal
-    ] >>= \ f -> f (m genCutReader) a b (pair <*> r <*> unit) (Run (uncurry ((. CutC.runCutA) . runReader)))
+    Cut.test (Hom (local (id @R))) (m genCutReader) a b (pair <*> r <*> unit) (Run (uncurry ((. CutC.runCutA) . runReader)))
   ] where
-  testMonad    run = Monad.test    (m gen) a b c (identity <*> unit) run
-  -- testMonadFix run = MonadFix.test (m gen) a b   (identity <*> unit) run
-  testCut      run = Cut.test      (m gen) a b   (identity <*> unit) run
+  testMonad    run = Monad.test        (m gen) a b c (identity <*> unit) run
+  -- testMonadFix run = MonadFix.test     (m gen) a b   (identity <*> unit) run
+  testCut      run = Cut.test (Hom id) (m gen) a b   (identity <*> unit) run
   genCutReader m = GenM $ \ a -> choice [runGenM (gen m) a, runGenM (Reader.gen r m) a]
-  testCutfailLocal (GenM m) a _ i (Run run) =
-    [ testProperty "cutfail commutes with local" (forall (i :. fn r :. m a :. Nil)
-      (\ i f m -> run ((local f cutfail <|> m) <$ i) === run (cutfail <$ i)))
-    ]
 
 
 gen :: (Has Cut sig m, Has NonDet sig m) => GenM m -> GenM m
@@ -52,20 +44,23 @@ gen (GenM m) = GenM $ \ a -> choice
   ]
 
 
+newtype Hom m = Hom (forall a . m a -> m a)
+
 test
   :: forall a b m f sig
   .  (Has Cut sig m, Has NonDet sig m, Arg a, Eq a, Eq b, Show a, Show b, Vary a, Functor f)
-  => GenM m
+  => Hom m
+  -> GenM m
   -> Gen a
   -> Gen b
   -> Gen (f ())
   -> Run f [] m
   -> [TestTree]
-test (GenM m) a b i (Run runCut)
+test (Hom hom) (GenM m) a b i (Run runCut)
   = testProperty "cutfail annihilates >>=" (forall (i :. fn @a (m a) :. Nil)
-    (\ i k -> runCut ((cutfail >>= k) <$ i) === runCut (cutfail <$ i)))
+    (\ i k -> runCut ((hom cutfail >>= k) <$ i) === runCut (hom cutfail <$ i)))
   : testProperty "cutfail annihilates <|>" (forall (i :. m a :. Nil)
-    (\ i m -> runCut ((cutfail <|> m) <$ i) === runCut (cutfail <$ i)))
+    (\ i m -> runCut ((hom cutfail <|> m) <$ i) === runCut (hom cutfail <$ i)))
   : testProperty "call delimits cutfail" (forall (i :. m a :. Nil)
-    (\ i m -> runCut ((call cutfail <|> m) <$ i) === runCut (m <$ i)))
+    (\ i m -> runCut ((hom (call (hom cutfail)) <|> m) <$ i) === runCut (m <$ i)))
   : NonDet.test (GenM m) a b i (Run runCut)
--- a/test/Cut.hs
+++ b/test/Cut.hs
@@ -55,11 +55,12 @@ test
   -> Gen (f ())
   -> Run f [] m
   -> [TestTree]
-test hom m a b i (Run runCut)
-  = testProperty "cutfail annihilates >>=" (forall (i :. fn @a (m a) :. Nil)
+test hom m = (\ a _ i (Run runCut) ->
+  [ testProperty "cutfail annihilates >>=" (forall (i :. fn @a (m a) :. Nil)
     (\ i k -> runCut ((hom cutfail >>= k) <$ i) === runCut (hom cutfail <$ i)))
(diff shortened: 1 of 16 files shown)


test/Cut.hs
```diff
@@ -22,34 +26,40 @@ tests = testGroup "Cut"
     [ testMonad
     -- , testMonadFix
     , testCut
-    ] >>= ($ RunL CutC.runCutA)
+    ] >>= ($ runL CutC.runCutA)
+  , testGroup "ReaderC · CutC" $
+    Cut.test (local (id @R)) (m (gen0 S.<> Reader.gen0 r) (\ m -> genN m S.<> Reader.genN r m)) a b (pair <*> r <*> unit) (Run (CutC.runCutA . uncurry runReader))
```
robrix: This is where things start to get really interesting.

We want to understand issues like #279 which only manifest due to relationships between different effects and carriers. In #279 specifically, it only happens when a `cutfail` occurs in a higher-order position in an effect run _after_ `CutC`. So `ReaderC r (CutC PureC)` doesn’t suffer this behaviour, because the `ReaderC` is eliminated first, but `CutC (ReaderC r PureC)` does.

In order to demonstrate this, we need to run the tests separately from `CutC PureC` to accommodate the different handlers and initial state. We also throw `Reader` actions into the generators for good measure, altho this is not technically required to manifest the bug. Finally, we provide a homomorphism for `Cut` actions—in this case `local id`—which the test suite applies to each action. (I’m not particularly thrilled with this approach, as it’s syntactically noisy and extremely manual, but I don’t have a better one at the moment and I wanted to get this in flight before fixing #279.)

test/Cut.hs
```diff
@@ -22,34 +26,40 @@ tests = testGroup "Cut"
...
+  -- , testGroup "CutC · ReaderC" $
+  --   Cut.test (local (id @R)) (m (gen0 S.<> Reader.gen0 r) (\ m -> genN m S.<> Reader.genN r m)) a b (pair <*> r <*> unit) (Run (uncurry ((. CutC.runCutA) . runReader)))
```
robrix: This is the failing test group; we omit it for the moment so that the PR will pass.

test/Cut.hs
```diff
@@ -22,34 +26,40 @@ tests = testGroup "Cut"
...
-    (\ m -> runCut (call cutfail <|> m) === runCut m))
-  : NonDet.test m a b (RunL runCut)
+test hom m a b i (Run runCut)
+  = testProperty "cutfail annihilates >>=" (forall (i :. fn @a (m a) :. Nil)
+    (\ i k -> runCut ((hom cutfail >>= k) <$ i) === runCut (hom cutfail <$ i)))
+  : testProperty "cutfail annihilates <|>" (forall (i :. m a :. Nil)
+    (\ i m -> runCut ((hom cutfail <|> m) <$ i) === runCut (hom cutfail <$ i)))
+  : testProperty "call delimits cutfail" (forall (i :. m a :. Nil)
+    (\ i m -> runCut ((hom (call (hom cutfail)) <|> m) <$ i) === runCut (m <$ i)))
+  : NonDet.test m a b i (Run runCut)
```
robrix: As described above, homomorphisms are applied to each action. One alternative I’m considering (but haven’t explored yet) is to define a polymorphic carrier which `hmap`s a function over all of the positions in an effect before delegating the effect to its underlying carrier.

test/Gen.hs
```diff
@@ -25,12 +26,9 @@ module Gen
 , identity
   -- * Handlers
 , Run(..)
-, type RunL
-, pattern RunL
-, type RunR
-, pattern RunR
-, type RunC
-, pattern RunC
```
robrix: Generalizing all of the tests over the input functor obviated the need for these.

test/Gen.hs
```diff
@@ -265,14 +232,18 @@ infixL p s f = Gen (pure (InfixL p s f))
...
 addLabel :: String -> Gen a -> Gen a
-addLabel s = Gen . (>>= \ a -> a <$ tell (Set.singleton (fromString s))) . runGen
+addLabel s = Gen . (>>= (<$ tell (Set.singleton (fromString s)))) . runGen
 
 
 data Term a where
   Pure :: (Int -> ShowS) -> a -> Term a
   InfixL :: Int -> String -> (a -> b -> c) -> Term (a -> b -> c)
   InfixR :: Int -> String -> (a -> b -> c) -> Term (a -> b -> c)
+  Pair :: Term (a -> b -> (a, b))
```
robrix: I got tired of pairs being shown as `(,) a b`.

test/NonDet.hs
```diff
@@ -24,29 +25,36 @@ tests = testGroup "NonDet"
...
-  ++ Choose.test m a b (RunL runNonDet)
+test m
+  = (\ a _ i (Run runNonDet) ->
+    [ testProperty "empty is the left identity of <|>"  (forall (i :. m a :. Nil)
+      (\ i m -> runNonDet ((empty <|> m) <$ i) === runNonDet (m <$ i)))
+    ,  testProperty "empty is the right identity of <|>" (forall (i :. m a :. Nil)
+      (\ i m -> runNonDet ((m <|> empty) <$ i) === runNonDet (m <$ i)))
+    ])
+  S.<> Empty.test  m
+  S.<> Choose.test m
```
robrix: I love the `Semigroup` instance for functions so much. Unfortunately, we still have to match `m` outside and pass it in specially because it’s a rank-n function.

test/Reader.hs
```diff
@@ -11,53 +12,60 @@ import qualified Control.Monad.Trans.Reader as ReaderT
...
 
 test
-  :: (Has (Reader r) sig m, Arg r, Eq a, Show a, Show r, Vary r)
+  :: (Has (Reader r) sig m, Arg r, Eq a, Show a, Show r, Vary r, Functor f)
   => Gen r
   -> GenM m
   -> Gen a
-  -> RunC r Identity m
+  -> Gen (f ())
+  -> Run (f :.: (,) r) Identity m
```
robrix: `Reader` already had input state, so we add an extra layer with `:.:`. Note that we have separate generators for the `f` part and the `(,) r` part, because we can’t in general extract `r` from `f (r, ())` without knowledge about what `f` is.

test/State.hs
```diff
@@ -25,45 +25,45 @@ tests = testGroup "State"
...
 
 test
   :: (Has (State s) sig m, Arg s, Eq a, Eq s, Show a, Show s, Vary s)
-  => Gen s
-  -> GenM m
+  => GenM m
   -> Gen a
-  -> RunC s ((,) s) m
+  -> Gen s
+  -> Run ((,) s) ((,) s) m
```
robrix: I haven’t yet generalized this one over the input state, because we don’t need it urgently. It can wait for some (interesting!) state/nondet tests, or something.

test/Cut.hs
```diff
@@ -22,34 +26,41 @@ tests = testGroup "Cut"
...
-    (\ m -> runCut (call cutfail <|> m) === runCut m))
-  : NonDet.test m a b (RunL runCut)
+test hom m = (\ a _ i (Run runCut) ->
+  [ testProperty "cutfail annihilates >>=" (forall (i :. fn @a (m a) :. Nil)
+    (\ i k -> runCut ((hom cutfail >>= k) <$ i) === runCut (hom cutfail <$ i)))
+  , testProperty "cutfail annihilates <|>" (forall (i :. m a :. Nil)
+    (\ i m -> runCut ((hom cutfail <|> m) <$ i) === runCut (hom cutfail <$ i)))
+  , testProperty "call delimits cutfail" (forall (i :. m a :. Nil)
+    (\ i m -> runCut ((hom (call (hom cutfail)) <|> m) <$ i) === runCut (m <$ i)))
+  ])
```
robrix: Reposting a comment that got hidden:

> As described above, homomorphisms are applied to each action. One alternative I’m considering (but haven’t explored yet) is to define a polymorphic carrier which `hmap`s a function over all of the positions in an effect before delegating the effect to its underlying carrier.

test/Reader.hs
```diff
@@ -11,53 +12,60 @@ import qualified Control.Monad.Trans.Reader as ReaderT
...
+  , testGroup "ReaderT"       $ testReader (runR (uncurry (flip ReaderT.runReaderT)    . lower))
+  , testGroup "RWST (Lazy)"   $ testReader (runR (uncurry (runRWST LazyRWST.runRWST)   . lower))
+  , testGroup "RWST (Strict)" $ testReader (runR (uncurry (runRWST StrictRWST.runRWST) . lower))
   ] where
-  testMonad    run = Monad.test    (m (gen r)) a b c (atom "(,)" (,) <*> r <*> unit) run
-  testMonadFix run = MonadFix.test (m (gen r)) a b   (atom "(,)" (,) <*> r <*> unit) run
-  testReader   run = Reader.test r (m (gen r)) a                                     run
+  testMonad    run = Monad.test    (m (gen0 r) (genN r)) a b c (Comp1 <$> (identity <*> (pair <*> r <*> unit))) run
+  testMonadFix run = MonadFix.test (m (gen0 r) (genN r)) a b   (Comp1 <$> (identity <*> (pair <*> r <*> unit))) run
+  testReader   run = Reader.test r (m (gen0 r) (genN r)) a                (identity <*>                 unit)   run
```
jkachmar: Huh, I didn't realize that the current module can implicitly be referred to as if it were a qualified import.

Am I understanding that correctly?

robrix: You are! Basically, every module in scope is implicitly available under its full name unless it’s renamed using `as`. And the current module is always in scope too. I imagine you can even make life difficult for yourself by importing other modules under the current module’s name.

robrix:
Alright the tests pass 🎉

PR #288, opened 2019-10-22
Shrink generated programs to subterms

robrix:
This PR allows us to shrink generated programs to subterms.

- [x] Depends on #283.

### DIFF
--- a/test/Catch.hs
+++ b/test/Catch.hs
@@ -16,7 +16,7 @@ genN
   -> GenM m
   -> GenTerm a
   -> [GenTerm (m a)]
-genN _ m a = [ addLabel "catchError" $ subtermM (m a) (\ m' -> infixL 9 "`catchError`" catchError <*> term m' <*> fn @e (m a)) ]
+genN _ m a = [ addLabel "catchError" $ subtermM (m a) (\ m' -> infixL 9 "`catchError`" catchError <*> m' <*> fn @e (m a)) ]
 
 
 test
--- a/test/Catch.hs
+++ b/test/Catch.hs
@@ -16,7 +16,7 @@ genN
   -> GenM m
   -> GenTerm a
   -> [GenTerm (m a)]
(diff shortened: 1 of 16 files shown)


test/Catch.hs
```diff
@@ -12,20 +12,20 @@ import Test.Tasty.Hedgehog
...
   .  (Has (Catch e) sig m, Arg e, Show e, Vary e)
-  => Gen e
+  => GenTerm e
   -> GenM m
-  -> Gen a
-  -> [Gen (m a)]
-genN _ m a = [ label "catchError" catchError <*> m a <*> fn @e (m a) ]
+  -> GenTerm a
+  -> [GenTerm (m a)]
+genN _ m a = [ addLabel "catchError" $ subtermM (m a) (\ m' -> infixL 9 "`catchError`" catchError <*> m' <*> fn @e (m a)) ]
```
robrix: Two things worth noting here:

1. We’re generating `catchError` as an infix term, so it’ll actually be shown as:

    ```haskell
    m `catchError` h
    ```

2. We use `subtermM` to generate the embedded action (but not the handler, because that would limit the distribution unnecessarily), which allows it to shrink to that action.

test/Gen.hs
```diff
@@ -220,23 +231,29 @@ showing = Pure . flip showsPrec <*> id
 showingFn :: (Show a, Show b) => Fn.Fn a b -> Term (a -> b)
 showingFn = Pure . flip showsPrec <*> Fn.apply
 
-atom :: String -> a -> Gen a
-atom s = Gen . pure . Pure (const (showString s))
 
-label :: String -> a -> Gen a
+type GenTerm = WriterT (Set.Set LabelName) Gen :.: Term
```
robrix: This turns out to do everything our hand-rolled `Gen` did, while also allowing us to easily wrap up `subtermM`/`subtermM2` 🎉

PR #292, opened 2019-10-23
Church.Cut: Fix cutfail in higher-order positions, & MonadFix

### LINKED ISSUE #279: Cut.Church: Higher-order effects allow backtracking past cutfail

### LINKED ISSUE #280: Cut.Church doesn’t obey the MonadFix/left-shrinking law

robrix:
- [x] Depends on #288.
- [x] Fixes #279.
- [x] Fixes #280.

### DIFF
--- a/src/Control/Carrier/Cut/Church.hs
+++ b/src/Control/Carrier/Cut/Church.hs
@@ -70,7 +70,7 @@ instance Fail.MonadFail m => Fail.MonadFail (CutC m) where
 -- | A single fixpoint is shared between all branches.
 instance MonadFix m => MonadFix (CutC m) where
   mfix f = CutC $ \ cons nil fail -> mfix
-    (runCut (fmap . (<|>) . pure) (pure empty) (pure cutfail) . f . run . runCut (<$) undefined undefined)
+    (runCut (fmap . (<|>) . pure) (pure empty) (pure cutfail) . f . run . runCut (<$) (error "mfix CutC: empty") (error "mfix CutC: cutfail"))
     >>= run . runCut (fmap . cons) (pure nil) (pure fail) where
   {-# INLINE mfix #-}
 
--- a/src/Control/Carrier/Cut/Church.hs
+++ b/src/Control/Carrier/Cut/Church.hs
@@ -67,14 +67,11 @@ instance Fail.MonadFail m => Fail.MonadFail (CutC m) where
   fail s = lift (Fail.fail s)
   {-# INLINE fail #-}
 
--- | Separate fixpoints are computed for each branch.
+-- | A single fixpoint is shared between all branches.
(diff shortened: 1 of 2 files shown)


src/Control/Carrier/Cut/Church.hs
```diff
@@ -92,5 +91,5 @@ instance (Algebra sig m, Effect sig) => Algebra (Cut :+: NonDet :+: sig) (CutC m
   eff (L (Call m k)) = CutC $ \ cons nil fail -> runCut (\ a as -> runCut cons as fail (k a)) nil nil m
   eff (R (L (L Empty)))      = empty
   eff (R (L (R (Choose k)))) = k True <|> k False
-  eff (R (R other))          = CutC $ \ cons nil _ -> eff (handle [()] (fmap concat . traverse runCutA) other) >>= foldr cons nil
+  eff (R (R other))          = CutC $ \ cons nil fail -> eff (handle (pure ()) (runCut (fmap . (<|>)) (pure empty) (pure cutfail)) other) >>= runCut cons nil fail
```
robrix: This trick worked out beautifully, and I intend to steal it for the other nondeterminism carriers in #286 (suitably repurposed): basically, we’re using `CutC m` as its _own_ state functor. If you’ll note, we use a very similar trick in the `MonadFix` instance, using `CutC PureC` as a representation of the fixpoint.

In both cases, this works precisely because it allows us to encode lists with soft and hard emptiness: i.e. lists differentiating between `empty` and `cutfail`. This means that we then propagate `cutfail`s through the fixpoint and higher-order actions correctly, instead of mapping them to `empty` and losing the essential information.

PR #295, opened 2019-10-24
Remove HFunctor in preparation for improvements to Effect

### LINKED ISSUE #235: What should the idiom for delegating the tail of the signature be?
`Carrier` instances often delegate the tail of the signature to the underlying carrier:

```haskell
newtype FooC m a = FooC { runFooC :: m a }

instance Carrier sig m => Carrier (Foo :+: sig) (FooC m) where
  eff (L …)     = …
  eff (R other) = FooC (eff (handleCoercible other))
```

when `FooC` is defined as a composition of other carriers, then this often also involves supplying a number of `R`s corresponding to the number of effects handled by the carriers it wraps:

```haskell
newtype FooC m a = FooC { runFooC :: ReaderC Int m a }

instance Carrier sig m => Carrier (Foo :+: sig) (FooC m) where
  eff (L …)     = …
  eff (R other) = FooC (eff (R (handleCoercible other)))
```

```haskell
newtype FooC m a = FooC { runFooC :: ReaderC Int (StateC String m) a }

instance Carrier sig m => Carrier (Foo :+: sig) (FooC m) where
  eff (L …)     = …
  eff (R other) = FooC (eff (R (R (handleCoercible other))))
```

and so on.

It turns out that `send` handles all of these uniformly: `send = alg . inj`, and `inj` abstracts over sums all the way up to a reflexive instance. This raises a few questions:

- What are the tradeoffs of using `send` vs. `eff` & some composition of `R`s? For example:
  - Is it ever ambiguous?
  - Does it harm inference?
  - Is it clearer as to the intent, or is it too magical?
  - Is it actually more convenient in practice?
  - Does it perform differently?
- Based on the above, should we change the default idiom used & documented in `fused-effects`?

robrix:
This PR:

- [x] Depends on #294.
- [x] Depends on #299.
- [x] Depends on #304.
- [x] 🔥s `HFunctor`.
- [x] Fixes #235: we subsume the `alg` into `handle…` functions. (See also #296 & #300.)


.gitignore
```diff
@@ -3,3 +3,4 @@
 dist-newstyle
 stack.yaml
 cabal.project.local
+.ghci
```
robrix: I’ve started using `cabal exec ghci` more and so I’m using this to set stuff up for that, but I don’t think we want to commit it since others will likely be using `cabal repl` and it might break or confuse things?

examples/ReinterpretLog.hs
```diff
@@ -179,7 +179,7 @@ instance
         unReinterpretLogC k
 
     R other ->
-      ReinterpretLogC (alg (R (handleCoercible other)))
+      ReinterpretLogC (handleCoercible other)
```
robrix: This is where I’ve come down re: #235: this is all boilerplate, so let’s at least make it minimal, obvious, & consistent boilerplate.

src/Control/Algebra.hs
```diff
@@ -61,17 +64,40 @@ instance Algebra Pure PureC where
...
+{-# INLINE handleIdentity #-}
+
+-- | Thread a 'Coercible' carrier through an 'Effect'.
+--
+-- This is applicable whenever @m@ is 'Coercible' to @n@, e.g. simple @newtype@s.
+--
+-- @since 1.0.0.0
+handleCoercible :: (Monad m, Effect eff, Member eff sig, Algebra sig n, Coercible m n) => eff m a -> n a
+handleCoercible = handleIdentity coerce
+{-# INLINE handleCoercible #-}
```
robrix: The new bit here is that `handleCoercible` calls (via `handleIdentity`) `send`, and thus no longer requires you to count `R`s on the left.

src/Control/Carrier/Error/Either.hs
```diff
@@ -51,6 +51,9 @@ instance (Alternative m, Monad m) => Alternative (ErrorC e m) where
 -- | 'ErrorC' passes 'MonadPlus' operations along to the underlying monad @m@, rather than combining errors à la 'ExceptT'.
 instance (Alternative m, Monad m) => MonadPlus (ErrorC e m)
 
-instance (Algebra sig m, Effect sig) => Algebra (Error e :+: sig) (ErrorC e m) where
-  alg = ErrorC . alg . handleCoercible
+instance Algebra sig m => Algebra (Error e :+: sig) (ErrorC e m) where
+  -- NB: 'send' (& thus 'handleCoercible') can’t send sums, so we decompose the sum manually.
+  alg (L (L op)) = ErrorC (handleCoercible op)
+  alg (L (R op)) = ErrorC (handleCoercible op)
+  alg (R op)     = ErrorC (handleCoercible op)
```
robrix: This is the one weird case: because `handleCoercible` uses `send`, and because `send` uses `Member`, and because `Member` can’t resolve sums of effects on the left, we have to decompose the sum here instead of doing just `alg = ErrorC . handleCoercible`.

I wouldn’t expect end users to have to do this very often, however, since it’s not very often you want to delegate _all_ of the effects you handle to something else (how I wish we could simply GND-derive the instance!), or that (as in this case), you’re handling multiple effects combined in a sum, let alone both.

src/Control/Carrier/Interpret.hs
```diff
@@ -54,19 +56,19 @@ reify a k = unsafeCoerce (Magic k) a
 --
 -- @since 1.0.0.0
 runInterpret
-  :: (HFunctor eff, Monad m)
+  :: (Effect eff, Monad m)
   => (forall x . eff m x -> m x)
   -> (forall s . Reifies s (Handler eff m) => InterpretC s eff m a)
   -> m a
-runInterpret f m = reify (Handler (InterpretC . f . handleCoercible)) (go m) where
+runInterpret f m = reify (Handler (InterpretC . fmap runIdentity . f . handle (Identity ()) (fmap Identity . coerce))) (go m) where
```
robrix: This is the other casualty of `handleCoercible`: we can’t use it here any more. I’m ok with that, again because this is something most users are unlikely to have to do themselves. In particular, users of `runInterpret` itself certainly don’t—`runInterpretState` bears no such scar.

src/Control/Effect/Catch/Internal.hs
```diff
@@ -13,8 +13,5 @@ data Catch e m k
 
 deriving instance Functor m => Functor (Catch e m)
 
-instance HFunctor (Catch e) where
-  hmap f (Catch m h k) = Catch (f m) (f . h) (f . k)
-
```
robrix: And now the good bit: 🔥

src/Control/Effect/Choose.hs
```diff
@@ -52,7 +52,7 @@ import qualified Data.Semigroup as S
 --
 -- @since 1.0.0.0
 (<|>) :: Has Choose sig m => m a -> m a -> m a
-(<|>) a b = send (Choose (bool b a))
+a <|> b = send (Choose (bool b a))
```
robrix: I really couldn’t tell you why I wrote it like that but when I noticed it I had to fix it.

robrix:
It’s pretty clear that we want #296, which means we can review & merge this. (#296 subsumes `HFunctor`, which is what makes this a reasonable thing to do.)

If it turns out that #296 is a no-go, we can revert this after the fact, but I really don’t think that’s likely.

PR #297, opened 2019-10-24
Reinterpreting effects docs

iand675:
Implements some docs for #9 

This is documentation is fairly narrative-oriented to try and motivate the process behind things. If this pseudo-blog-post style of documentation isn't quite right I could pull it out into a GitHub repo of its own or something.

### DIFF
--- a/docs/reinterpreting_effects.md
+++ b/docs/reinterpreting_effects.md
@@ -1,6 +1,6 @@
 # Reinterpreting effects
 
-This article assumes you are already familiar with [defining effects and their handlers](https://github.com/fused-effects/fused-effects#defining-new-effects).
+This article assumes you are already familiar with [defining effects and their handlers](defining_effects.md).
 
 One of the nice aspects of effects is that they can support multiple effect handlers. Effects only specify actions, they don't actually perform them. Therefore, it's possible to "reinterpret" effects. There are multiple senses in which an effect can be reinterpreted:
 
@@ -9,7 +9,7 @@ One of the nice aspects of effects is that they can support multiple effect hand
 
 Let's explore both of these effect interpretation strategies with a small motivating example:
 
-> We would like to implement a client library for an [HTTP-based API that provides interesting cat facts](https://alexwohlbruck.github.io/cat-facts/docs/).
+✨ We would like to implement a client library for an [HTTP-based API that provides interesting cat facts](https://alexwohlbruck.github.io/cat-facts/docs/). ✨
 
 Let's break down some of the properties of the API client that would be desirable for a production use case:
 
@@ -91,7 +91,7 @@ sendRequest :: Has Http sig m => HTTP.Request -> m (HTTP.Response L.ByteString)
 sendRequest r = send (SendRequest r pure)
 ```
 
-As you can hopefully see, we've decomposed the original problem into several small effects that take care of a different layer of the original problem description. Now let's take a look at how we can compose them together to achieve our original goals.
+The listFacts function provides the ‘what’ of this API, and the sendRequest function provides the ‘how’. In decomposing this problem into a set of effects, each responsible for a single layer of the original problem description, we provide ourselves with a flexible, composable vocabulary rather than a single monolithic action.
 
 ## "Stacking" effects
 
@@ -124,10 +124,7 @@ will actually return JSON, so for this implementation we have to account for fai
 a great opportunity to see how effect handlers can actually rely on *multiple underlying effects*!
 
 ``` haskell
-runCatFactsApi :: CatFactsApiC m a -> m a
-runCatFactsApi = runCatFactsApiC
-
-newtype CatFactsApiC m a = CatFactsApiC { runCatFactsApiC :: m a }
+newtype CatFactsApi m a = CatFactsApi { runCatFactsApi :: m a }
  deriving newtype
       ( Monad
       , Functor
@@ -139,16 +136,18 @@ newtype CatFactsApiC m a = CatFactsApiC { runCatFactsApiC :: m a }
 catFactsEndpoint :: HTTP.Request
 catFactsEndpoint = HTTP.parseRequest_ "https://cat-fact.herokuapp.com/facts/random"
 
-instance ( Has (Http :+: Throw JsonParseError :+: Throw InvalidContentType) sig m
+instance ( Has Http sig m
+         , Has (Throw JsonParseError) sig m
+         , Has (Throw InvalidContentType) sig m
          , Algebra sig m
          ) =>
-         Algebra (CatFactClient :+: sig) (CatFactsApiC m) where
+         Algebra (CatFactClient :+: sig) (CatFactsApi m) where
   eff (L (ListFacts numberOfFacts k)) = do
     resp <- sendRequest (catFactsEndpoint { HTTP.queryString = "?amount=" <> B.pack (show numberOfFacts) })
     case lookup hContentType (HTTP.responseHeaders resp) of
       Just "application/json; charset=utf-8" -> decodeOrThrow (HTTP.responseBody resp) >>= k
       other -> throwError (InvalidContentType (show other))
-  eff (R other) = CatFactsApiC (eff (handleCoercible other))
+  eff (R other) = CatFactsApi (eff (handleCoercible other))
 ```
 
 We implement a `CatFacts` effect handler that depends on _three_ underlying effects:
@@ -162,10 +161,8 @@ The nice aspect of this is that we have neatly contained the failure scenarios t
 Now we need to support performing HTTP requests:
 
 ``` haskell
-runHttp :: HttpC m a -> m a
-runHttp = runHttpC
 
-newtype HttpC m a = HttpC { runHttpC :: m a }
+newtype HttpClient m a = HttpClient { runHttp :: m a }
  deriving newtype
       ( Monad
       , Functor
@@ -174,14 +171,14 @@ newtype HttpC m a = HttpC { runHttpC :: m a }
       , Alternative
       )
 
-instance (MonadIO m, Algebra sig m) => Algebra (Http :+: sig) (HttpC m) where
+instance (MonadIO m, Algebra sig m) => Algebra (Http :+: sig) (HttpClient m) where
   eff (L (SendRequest req k)) = liftIO (HTTP.getGlobalManager >>= HTTP.httpLbs req) >>= k
-  eff (R other) = HttpC (eff (handleCoercible other))
+  eff (R other) = HttpClient (eff (handleCoercible other))
 ```
 
-Note for the above code snippets how the `CatFactsApiC` carrier delegates fetching JSON to any other effect that supports retrieving JSON given an HTTP request specification.
+Note for the above code snippets how the `CatFactsApi` carrier delegates fetching JSON to any other effect that supports retrieving JSON given an HTTP request specification.
 
-Note as well that `CatFactsApiC` itself doesn't know how to perform an HTTP request. It delegates the request itself to a handler that implements the `Algebra` class for `(Http :+: sig)`.
+Note as well that `CatFactsApi` itself doesn't know how to perform an HTTP request. It delegates the request itself to a handler that implements the `Algebra` class for `(Http :+: sig)`.
 
 Putting it all together for the actual production use case:
 
@@ -230,7 +227,7 @@ to surface from least-pure parts of code. In this case, we should therefore impl
 to experiment with the most failure-prone area: the network itself.
 
 ``` haskell
-newtype MockHttpC m a = MockHttpC { runMockHttpC :: ReaderC (HTTP.Request -> IO (HTTP.Response L.ByteString)) m a }
+newtype MockHttpClient m a = MockHttpClient { runMockHttpClient :: ReaderC (HTTP.Request -> IO (HTTP.Response L.ByteString)) m a }
  deriving newtype
       ( Monad
       , Functor
@@ -240,11 +237,11 @@ newtype MockHttpC m a = MockHttpC { runMockHttpC :: ReaderC (HTTP.Request -> IO 
       )
 
 runMockHttp :: (HTTP.Request -> IO (HTTP.Response L.ByteString)) -> MockHttpC m a -> m a
-runMockHttp responder m = runReader responder (runMockHttpC m)
+runMockHttp responder m = runReader responder (runMockHttpClient m)
 
-instance (MonadIO m, Algebra sig m) => Algebra (Http :+: sig) (MockHttpC m) where
-  eff (L (SendRequest req k)) = MockHttpC ask >>= \responder -> liftIO (responder req) >>= k
-  eff (R other) = MockHttpC (eff (R (handleCoercible other)))
+instance (MonadIO m, Algebra sig m) => Algebra (Http :+: sig) (MockHttpClient m) where
+  eff (L (SendRequest req k)) = MockHttpClient ask >>= \responder -> liftIO (responder req) >>= k
+  eff (R other) = MockHttpClient (eff (R (handleCoercible other)))
 
 faultyNetwork :: HTTP.Request -> IO (HTTP.Response L.ByteString)
 faultyNetwork req = throwIO (HTTP.HttpExceptionRequest req HTTP.ConnectionTimeout)
--- a/docs/reinterpreting_effects.md
+++ b/docs/reinterpreting_effects.md
@@ -13,18 +13,47 @@ Let's explore both of these effect interpretation strategies with a small motiva
 
 Let's break down some of the properties of the API client that would be desirable for a production use case:
 
-- We would like to have our cat facts API be able to support different cat fact data sources in the future.
-- We would like to be able to mock failure conditions (such as network connectivity issues) for testing purposes.
-- We would like to be able to track timing metrics for how quickly we can retrieve cat facts.
+1. We would like to have our cat facts API be able to support different cat fact data sources in the future.
+2. We would like to be able to mock failure conditions (such as network connectivity issues) for testing purposes.
+3. We would like to be able to track timing metrics for how quickly we can retrieve cat facts.
 
 ### Initial setup
 
 ``` haskell
-module CatFacts where
-import Data.Aeson -- from the aeson package
-import Data.Time
-import qualified Network.HTTP.Client as HTTP -- from the http-client package
-import qualified Network.HTTP.Client.TLS as HTTP -- from the http-client-tls package
+{-# LANGUAGE ExistentialQuantification, DeriveAnyClass, DeriveFunctor,
+DeriveGeneric, DerivingStrategies, FlexibleContexts, FlexibleInstances,
+GeneralizedNewtypeDeriving, OverloadedStrings, MultiParamTypeClasses,
+RankNTypes, TypeApplications, TypeOperators, UndecidableInstances #-}
+module CatFacts
+    ( main
+    ) where
+-- from base
+import Control.Applicative
+import Control.Exception (throwIO)
+import GHC.Generics (Generic1)
+-- from fused-effects
+import Control.Algebra
+import Control.Carrier.Reader
+import Control.Carrier.Error.Either
+import Control.Carrier.Interpret
+-- from transformers
+import Control.Monad.IO.Class
+-- From aeson
+import Data.Aeson
+-- From bytestring
+import qualified Data.ByteString.Char8 as B
+import qualified Data.ByteString.Lazy as L
+-- From time
+import Data.Time.Clock
+-- From http-client
+import qualified Network.HTTP.Client as HTTP
+import Network.HTTP.Client.Internal (Response(..), ResponseClose(..))
+-- From http-client-tls
+import qualified Network.HTTP.Client.TLS as HTTP
+-- From http-status
+import Network.HTTP.Types.Header
+import Network.HTTP.Types.Status
+import Network.HTTP.Types.Version
 ```
 
 Since one of the best parts about effects is being able to think at a domain language level,
@@ -37,6 +66,9 @@ data CatFact = CatFact
   { catFact :: String
   } deriving (Show)
 
+instance FromJSON CatFact where
+  parseJSON = withObject "CatFact" (\o -> CatFact <$> o .: "text")
+
 -- | Our high level effect type that will be able to target different data sources.
 data CatFactClient m k
   = ListFacts Int {- ^ Number of facts to fetch -} ([CatFact] -> m k)
@@ -47,17 +79,7 @@ listFacts :: Has CatFactClient sig m => Int -> m [CatFact]
 listFacts n = send (ListFacts n pure)
 ```
 
-Now that we have our very simple DSL in place, let's think about the underlying API: we know that it's JSON-based, so let's introduce the notion of a handler that is provided a request and hands back something that can be parsed out from a JSON document.
-
-``` haskell
-data JsonFetch req m k
-  = forall a. (FromJSON a) => JsonFetch req (a -> m k)
-
-fetchJson :: (Has (JsonFetch req) sig m, FromJSON a) => req -> m a
-fetchJson r = send (JsonFetch r pure)
-```
-
-Let's introduce an effect type for making arbitrary requests over HTTP:
+Now that we have our very simple DSL in place, let's think about the underlying API: we know that it's an HTTP-based system, so let's introduce the notion of a handler that is provided a request and hands back an HTTP response.
 
 ``` haskell
 data Http m k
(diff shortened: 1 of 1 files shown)


docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,362 @@
...
+This article assumes you are already familiar with [defining effects and their handlers](https://github.com/fused-effects/fused-effects#defining-new-effects).
+
+One of the nice aspects of effects is that they can support multiple effect handlers. Effects only specify actions, they don't actually perform them. Therefore, it's possible to "reinterpret" effects. There are multiple senses in which an effect can be reinterpreted:
+
+- Implementing an effect in terms of other effects. "Stacking" effects is a powerful tool for cleanly dividing implementations into the relevant abstraction layers with minimal leakage of implementation details.
+- Rewriting an effect and/or performing actions with the effect value and then performing the originally intended effect. This technique is conceptually similar to the middleware pattern commonly used in web applications.
+
+Let's explore both of these effect interpretation strategies with a small motivating example:
+
+> We would like to implement a client library for an [HTTP-based API that provides interesting cat facts](https://alexwohlbruck.github.io/cat-facts/docs/).
```
patrickt: Stylistic nit: no need for this to be blockquoted; merge it with the previous sentence and it’ll be grand.

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,362 @@
...
+import Network.HTTP.Types.Version
+```
+
+Since one of the best parts about effects is being able to think at a domain language level,
+let's start with defining the desired data which we wish to retrieve and an interface that
+feels natural to work with:
+
+``` haskell
+-- | The basic fact that we will retrieve.
+data CatFact = CatFact
```
patrickt:
```suggestion
newtype CatFact = CatFact
```

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,362 @@
...
+data Http m k
+  = SendRequest HTTP.Request (HTTP.Response L.ByteString -> m k)
+  deriving stock (Functor, Generic1)
+  deriving anyclass (HFunctor, Effect)
+
+sendRequest :: Has Http sig m => HTTP.Request -> m (HTTP.Response L.ByteString)
+sendRequest r = send (SendRequest r pure)
+```
+
+As you can hopefully see, we've decomposed the original problem into several small effects that take care of a different layer of the original problem description. Now let's take a look at how we can compose them together to achieve our original goals.
```
patrickt: I’d replace "As you can hopefully see” with something stronger like “The `listFacts` function provides the ‘what’ of this API, and the `sendRequest` function provides the ‘how’. In decomposing this problem into a set of effects, each responsible for a single layer of the original problem description, we provide ourselves with a flexible, composable vocabulary rather than a single monolithic action.”

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,362 @@
...
+
+## "Stacking" effects
+
+### The production use-case
+
+Now that we have these 2 mini-DSL effect types established, we need to stitch them together.
```
patrickt:
```suggestion
Now that we have these two mini-DSL effect types established, we need to stitch them together.
```

docs/reinterpreting_effects.md
patrickt: Strictly speaking, this is more correct, and it’s example code so we should be on our best behavior.

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,362 @@
...
+
+Let's take a moment to think about what could go wrong with an HTTP API from which we plan to fetch some JSON and convert it into a list of `CatFact`s.
+
+We can conceive that the server might occasionally return a malformed JSON response:
+
+``` haskell
+data JsonParseError = JsonParseError String
```
patrickt:
```suggestion
newtype JsonParseError = JsonParseError String
```

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,362 @@
...
+  deriving (Show, Eq)
+
+decodeOrThrow :: (Has (Throw JsonParseError) sig m, FromJSON a) => L.ByteString -> m a
+decodeOrThrow = either (throwError . JsonParseError) pure . eitherDecode
+```
+
+A more HTTP-centric issue is that we might receive a content type we can't use. In this case, anything that's not `application/json`:
+
+``` haskell
+data InvalidContentType = InvalidContentType String
```
patrickt:
```suggestion
newtype InvalidContentType = InvalidContentType String
```

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,362 @@
...
+2. `Throw JsonParseError` - we need to be able to signal that some aspect of the JSON wasn't what we expected.
+3. `Throw InvalidContentType` - we need to be able to signal what we got wasn't JSON at all!
+
+The nice aspect of this is that we have neatly contained the failure scenarios to their relevant strata rather than leaking them into the higher-level abstraction (`listFacts`)!
+
+Now we need to support performing HTTP requests:
+
+``` haskell
+runHttp :: HttpC m a -> m a
+runHttp = runHttpC
```
patrickt: We’ve taken to avoiding `C` suffixes if possible, so I’d recommend axing this and changing the newtype’s record selector to `runHttp`.

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,362 @@
...
+Putting it all together for the actual production use case:
+
+``` haskell
+handlePrint :: Either InvalidContentType (Either JsonParseError [CatFact]) -> IO ()
+handlePrint r =
+  case r of
+    Left invalidContentTypeError -> print invalidContentTypeError
+    Right ok -> case ok of
+      Left jsonParseError -> print jsonParseError
+      Right facts -> mapM_ (putStrLn . catFact) facts
```
patrickt:
```suggestion
      Right facts -> traverse (putStrLn . catFact) facts
```

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,362 @@
+# Reinterpreting effects
+
+This article assumes you are already familiar with [defining effects and their handlers](https://github.com/fused-effects/fused-effects#defining-new-effects).
```
patrickt: You probably want to link to the `defining_effects.md` in this folder.

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,362 @@
...
+
+Produces:
+
+```
+The Bengal is the result of crossbreeding between domestic cats and Asian leopard cats, and its name is derived from the scientific name for the Asian leopard cat (Felis bengalensis).
+A happy cat holds her tail high and steady.
+Kittens remain with their mother till the age of 9 weeks.
+Recent studies have shown that cats can see blue and green. There is disagreement as to whether they can see red.
+A steady diet of dog food may cause blindness in your cat - it lacks taurine.
+Cat owners are 25% likely to pick George Harrison as their favorite Beatle.
```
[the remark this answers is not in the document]
iand675: Could be? I think the facts are crowdsourced, so hard to say.

iand675:
Believe I've addressed all of the reviews. 🚀

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
+# Reinterpreting effects
+
+This article assumes you are already familiar with [defining effects and their handlers](defining_effects.md).
+
+One of the nice aspects of effects is that they can support multiple effect handlers. Effects only specify actions, they don't actually perform them. Therefore, it's possible to "reinterpret" effects. There are multiple senses in which an effect can be reinterpreted:
+
+- Implementing an effect in terms of other effects. "Stacking" effects is a powerful tool for cleanly dividing implementations into the relevant abstraction layers with minimal leakage of implementation details.
```
robrix: This is what I typically think of as “reinterpreting;” I think I picked up that specific nomenclature from @lexi-lambda’s `freer-simple`.

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+- Rewriting an effect and/or performing actions with the effect value and then performing the originally intended effect. This technique is conceptually similar to the middleware pattern commonly used in web applications.
```
robrix: 1. We (and others) usually refer to this as “interposition,” following Oleg Kiselyov et al’s work on extensible effects.

2. Middleware is an excellent point of comparison ✨ I’ve typically thought of aspect-oriented programming, but that’s much less likely to be familiar, so I’mma borrow this example in future 😊

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+
+Let's explore both of these effect interpretation strategies with a small motivating example:
+
+✨ We would like to implement a client library for an [HTTP-based API that provides interesting cat facts](https://alexwohlbruck.github.io/cat-facts/docs/). ✨
+
+Let's break down some of the properties of the API client that would be desirable for a production use case:
+
+1. We would like to have our cat facts API be able to support different cat fact data sources in the future.
+2. We would like to be able to mock failure conditions (such as network connectivity issues) for testing purposes.
+3. We would like to be able to track timing metrics for how quickly we can retrieve cat facts.
```
robrix: This is just a superb motivating example: at once cute, simple, and a lucid and relevant demonstration of real-world engineering problems.

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+-- From http-client
+import qualified Network.HTTP.Client as HTTP
+import Network.HTTP.Client.Internal (Response(..), ResponseClose(..))
+-- From http-client-tls
+import qualified Network.HTTP.Client.TLS as HTTP
+-- From http-status
+import Network.HTTP.Types.Header
+import Network.HTTP.Types.Status
+import Network.HTTP.Types.Version
+```
```
robrix: This makes me wonder if we should rewrite this as a literate Haskell file and compile it into the `examples` component (with or without any unit tests—I’m not fussed) so it doesn’t go stale. On the other hand, you’ve already done a ton of work writing this and I don’t want to make the process any more arduous than I already have with my concurrent changes 😂

What do you think?

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+
+Since one of the best parts about effects is being able to think at a domain language level,
+let's start with defining the desired data which we wish to retrieve and an interface that
+feels natural to work with:
```
robrix: 🙏 Soft wrap, please.

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+  } deriving (Show)
+
+instance FromJSON CatFact where
+  parseJSON = withObject "CatFact" (\o -> CatFact <$> o .: "text")
+
+-- | Our high level effect type that will be able to target different data sources.
+data CatFactClient m k
+  = ListFacts Int {- ^ Number of facts to fetch -} ([CatFact] -> m k)
+  deriving stock (Functor, Generic1)
+  deriving anyclass (HFunctor, Effect)
```
robrix: 1. `HFunctor` is gone as of #295.

2. I’d strongly prefer to avoid `DeriveAnyClass` & `DerivingStrategies` since I think they present an unnecessarily complicated picture of the use of the library; `instance Effect` will suffice.

(Ditto for the other effects you’ve defined, naturally.)

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+Now that we have our very simple DSL in place, let's think about the underlying API: we know that it's an HTTP-based system, so let's introduce the notion of a handler that is provided a request and hands back an HTTP response.
+
+``` haskell
+data Http m k
+  = SendRequest HTTP.Request (HTTP.Response L.ByteString -> m k)
+  deriving stock (Functor, Generic1)
+  deriving anyclass (HFunctor, Effect)
+
+sendRequest :: Has Http sig m => HTTP.Request -> m (HTTP.Response L.ByteString)
+sendRequest r = send (SendRequest r pure)
```
robrix: Yeah so this is pretty dang cool! Makes me wonder if we should have a `fused-effects-http` package or something.

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+```
+
+The listFacts function provides the ‘what’ of this API, and the sendRequest function provides the ‘how’. In decomposing this problem into a set of effects, each responsible for a single layer of the original problem description, we provide ourselves with a flexible, composable vocabulary rather than a single monolithic action.
```
robrix:
```suggestion
The `listFacts` function provides the ‘what’ of this API, and the sendRequest function provides the ‘how’. In decomposing this problem into a set of effects, each responsible for a single layer of the original problem description, we provide ourselves with a flexible, composable vocabulary rather than a single monolithic action.
```

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+decodeOrThrow :: (Has (Throw JsonParseError) sig m, FromJSON a) => L.ByteString -> m a
+decodeOrThrow = either (throwError . JsonParseError) pure . eitherDecode
+```
+
+A more HTTP-centric issue is that we might receive a content type we can't use. In this case, anything that's not `application/json`:
+
+``` haskell
+newtype InvalidContentType = InvalidContentType String
+  deriving (Show, Eq)
+
```
robrix:
```suggestion
```

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+```
+
+Now we need to support fetching JSON given an HTTP request. We have no guarantee that an arbitrary HTTP request
+will actually return JSON, so for this implementation we have to account for failure conditions. This provides
+a great opportunity to see how effect handlers can actually rely on *multiple underlying effects*!
```
robrix: 🙏 Soft wrap, please.

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+catFactsEndpoint :: HTTP.Request
+catFactsEndpoint = HTTP.parseRequest_ "https://cat-fact.herokuapp.com/facts/random"
+
+instance ( Has Http sig m
+         , Has (Throw JsonParseError) sig m
+         , Has (Throw InvalidContentType) sig m
+         , Algebra sig m
+         ) =>
+         Algebra (CatFactClient :+: sig) (CatFactsApi m) where
+  eff (L (ListFacts numberOfFacts k)) = do
```
robrix:
```suggestion
  alg (L (ListFacts numberOfFacts k)) = do
```

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+    resp <- sendRequest (catFactsEndpoint { HTTP.queryString = "?amount=" <> B.pack (show numberOfFacts) })
+    case lookup hContentType (HTTP.responseHeaders resp) of
+      Just "application/json; charset=utf-8" -> decodeOrThrow (HTTP.responseBody resp) >>= k
+      other -> throwError (InvalidContentType (show other))
+  eff (R other) = CatFactsApi (eff (handleCoercible other))
```
robrix:
```suggestion
  alg (R other) = CatFactsApi (handleCoercible other)
```

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+1. `Http` - we need to be able to make requests
+2. `Throw JsonParseError` - we need to be able to signal that some aspect of the JSON wasn't what we expected.
+3. `Throw InvalidContentType` - we need to be able to signal what we got wasn't JSON at all!
+
+The nice aspect of this is that we have neatly contained the failure scenarios to their relevant strata rather than leaking them into the higher-level abstraction (`listFacts`)!
+
+Now we need to support performing HTTP requests:
+
+``` haskell
+
```
robrix:
```suggestion
```

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+  eff (R other) = CatFactsApi (eff (handleCoercible other))
+```
+
+We implement a `CatFacts` effect handler that depends on _three_ underlying effects:
+
+1. `Http` - we need to be able to make requests
+2. `Throw JsonParseError` - we need to be able to signal that some aspect of the JSON wasn't what we expected.
+3. `Throw InvalidContentType` - we need to be able to signal what we got wasn't JSON at all!
+
+The nice aspect of this is that we have neatly contained the failure scenarios to their relevant strata rather than leaking them into the higher-level abstraction (`listFacts`)!
```
robrix: This is really lovely!

There’s another possible implementation that might be preferable in some situations, which is to define the `CatFactsApi` carrier as a wrapper for some or all of the required carriers, e.g.:

```haskell
newtype CatFactsApi m a = CatFactsApi { runCatFactsApi :: ThrowC JsonParseError (ThrowC InvalidContentType (HttpClient m)) a }
```

If we do this, we don’t need the `Has` constraints, but on the other hand we might be making redundant work for ourselves if other stuff we use can also throw those errors, and we’d absolutely be limiting our ability to mock out the http client. This might be worth mentioning in passing, because the tradeoffs are super interesting and relevant when designing carriers; but I can also see an argument for pushing that to a future PR/document. What do you think?

docs/reinterpreting_effects.md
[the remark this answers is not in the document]
robrix: Anecdata, @patrickt!

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+```
+
+### Testing with alternative effect handlers
+
+Per point 2. of our initial implementation criteria, we want to be able to simulate failure cases for testing purposes. This is a great
+case for swapping in an alternative effect handler for our HTTP layer.
+
+This time let's go from the bottom up. In situations where IO is involved, failure scenarios tend
+to surface from least-pure parts of code. In this case, we should therefore implement some facilities
+to experiment with the most failure-prone area: the network itself.
```
robrix: 🙏 Soft wrap, please.

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+      , Applicative
+      , MonadIO
+      , Alternative
+      )
+
+runMockHttp :: (HTTP.Request -> IO (HTTP.Response L.ByteString)) -> MockHttpC m a -> m a
+runMockHttp responder m = runReader responder (runMockHttpClient m)
+
+instance (MonadIO m, Algebra sig m) => Algebra (Http :+: sig) (MockHttpClient m) where
+  eff (L (SendRequest req k)) = MockHttpClient ask >>= \responder -> liftIO (responder req) >>= k
```
robrix:
```suggestion
  alg (L (SendRequest req k)) = MockHttpClient ask >>= \responder -> liftIO (responder req) >>= k
```

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+  eff (R other) = MockHttpClient (eff (R (handleCoercible other)))
```
robrix:
```suggestion
  alg (R other) = MockHttpClient (handleCoercible other)
```

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+ deriving newtype
+      ( Monad
+      , Functor
+      , Applicative
+      , MonadIO
+      , Alternative
+      )
+
+instance (MonadIO m, Algebra sig m) => Algebra (Http :+: sig) (HttpClient m) where
+  eff (L (SendRequest req k)) = liftIO (HTTP.getGlobalManager >>= HTTP.httpLbs req) >>= k
```
robrix:
```suggestion
  alg (L (SendRequest req k)) = liftIO (HTTP.getGlobalManager >>= HTTP.httpLbs req) >>= k
```

This is a great example of working at the “right” level of abstraction—lifting the HTTP actions into `HttpClient` vs. putting everything inside `HttpClient $ …` and therefore having to do `>>= runHttp . k` at the end there. I really dig this.

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+  eff (R other) = HttpClient (eff (handleCoercible other))
```
robrix:
```suggestion
  alg (R other) = HttpClient (handleCoercible other)
```

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+badJson req = pure resp
+  where
+    resp = Response
+      { responseStatus = ok200
+      , responseVersion = http11
+      , responseHeaders = [("Content-Type", "application/json; charset=utf-8")]
+      , responseBody = "{}"
+      , responseCookieJar = mempty
+      , responseClose' = ResponseClose (pure ())
+      }
```
robrix: This is _super_ nitpicky, so feel free to ignore, but I don’t feel like the `where` clause is buying us much. We could simplify these to e.g.:

```haskell
badJson req = pure Response
  { responseStatus = ok200
  , …
  }
```

due to the unreasonably high precedence of record syntax.

Meme of southern preacher exhorting his congregation, captioned “the precedence of record update syntax is TOO DAMN HIGH”

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+
+Building new effect handling algebras can be a little bit verbose. In simpler situations, we may want to simply operate
+on an effect without having to implement a whole new `Algebra` instance. We still have yet to build a solution to tracking
+operational metrics (like request timings), so let's look at how to build a sort of "effect middleware" using `InterpretC`.
+
+`InterpretC` is an effect carrier that is intended for prototyping new effects that passes a callback function each
+occurence of the specified effect type that is called via `send`. One trick that can be useful is to intercept an effect,
+operate on the effect, and then re-`send` the effect. In other words, it's perfectly valid to have multiple handlers
+for the same effect type and dispatch to the ones higher in the effect stack! Let's use this approach to time and log
+our HTTP requests:
```
robrix: Soft wrap, please 🙏

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+
+``` haskell
+main :: IO ()
+main = runHttp (traceHttp catFactsRunner) >>= handlePrint
+```
+
+Returns:
+
+```
+/facts/random ... [200] took 0.979107082s
```
robrix: _Beautiful_. Wow.

docs/reinterpreting_effects.md
```diff
@@ -0,0 +1,359 @@
...
+### Conclusion
+
+Reviewing our initial criteria, we have an eminently extensible system that lets us maintain a healthy separation of concerns--
+All while still allowing non-invasive behavior changes through the ability to intercept, rewrite, and resend effects!
+
+- [x] We would like to have our cat facts API be able to support different cat fact data sources in the future.
+- [x] We would like to be able to mock failure conditions (such as network connectivity issues) for testing purposes.
+- [x] We would like to be able to track timing metrics for how quickly we can retrieve cat facts.
+
+![Mission accomplished](https://media.giphy.com/media/l6Td5sKDNmDGU/giphy.gif)
```
robrix: This gif is the icing on the cake ✨

robrix:
(Also, sorry for the delay in reviewing! I wanted to be able to give it the attention it deserves, and I’m very glad I did—it was well worth a careful reading 💖)

robrix:
> I’ve proposed moving this to a literate haskell file in `examples`, but upon reflection, that might be a more substantial change than is really reasonable right now, so feel free to decline or punt to a future PR; I would like (more of) our docs to be automatically verified, but that certainly doesn’t have to fall on you!

Slight correction here: the choice to move this into `examples` is separate from the choice to write it using literate Haskell. We don’t need to do both.

docs/reinterpreting_effects.md
iand675: I'm not totally opposed to it, but it makes me sad that literate Haskell loses formatting options. What will I do without my GIFs?

docs/reinterpreting_effects.md
iand675: I definitely think so. One of my bigger motivations for integrating `fused-effects` into future work is the ability to instrument effects, so any common side-effects like this or DB access would ideally be shareable between library implementors IMO. That being said, it has the potential to devolve a bit quickly in terms of granularity when considering how to support different HTTP client libs (`curl`? `http-client`? `http2-client`?). 

Plus, in this case I chose to constrain the response content reader to be a lazy bytestring. It might take a bit of thought to play nice with `conduit` et al.

docs/reinterpreting_effects.md
iand675: TIL

docs/reinterpreting_effects.md
iand675: Hm, I feel like this might merit its own post. I'm not convinced that forcing the implementation to exclusively use `ThrowC` as the only carrier for `Throw` in this case is ideal. This document is focused on promoting effect reinterpretation, so this seems to run counter to that point and might cause confusion.

It might be the case though that I'm missing some benefits of the tradeoff beyond eliding some `Has` constraints?

docs/reinterpreting_effects.md
iand675: Thanks, I'm flattered! Although I'm actually not sure what part of this is merits this response 😄

docs/reinterpreting_effects.md
robrix: 😆

robrix: (See also #310 for an alternative!)

docs/reinterpreting_effects.md
robrix: @iand675 abstraction, basically—you can completely encapsulate the underlying action. Think of `Control.Carrier.Writer.Strict.WriterC`, which is implemented as a wrapper around `StateC`—if it were implemented using a `Has` constraint instead, then it would be easy for another action to `put mempty`, violating `Writer`’s invariants. Reinterpreting-by-wrapping allows us to use `StateC`—and even the `State` operations—without having to expose the `State w` effect to the rest of the program.

All of which is to say, I think you’re exactly right that this deserves its own post 👍

docs/reinterpreting_effects.md
robrix: Layering on timing like this was a really sweet solution ✨

robrix:
professor farnsworth – evil laughter lightning

PR #299, opened 2019-10-24
Acyclic module graph

robrix:
This PR removes all the cyclic imports by moving the effect datatypes for which we’ve defined `Algebra` instances in `Control.Algebra` (i.e. algebras for `base` or `transformers` types) into internal modules.

- [x] Depends on #294.


fused-effects.cabal
```diff
@@ -84,6 +84,17 @@ library
...
+    Control.Effect.Catch.Internal
+    Control.Effect.Choose.Internal
+    Control.Effect.Empty.Internal
+    Control.Effect.Error.Internal
+    Control.Effect.Lift.Internal
+    Control.Effect.NonDet.Internal
+    Control.Effect.Reader.Internal
+    Control.Effect.State.Internal
+    Control.Effect.Throw.Internal
+    Control.Effect.Writer.Internal
```
robrix: We don’t export any of these modules, because they don’t define any functionality at all, only datatypes which are re-exported by the public interfaces.

src/Control/Algebra.hs
```diff
@@ -52,6 +52,11 @@ class (HFunctor sig, Monad m) => Algebra sig m | m -> sig where
   -- | Construct a value in the carrier for an effect signature (typically a sum of a handled effect and any remaining effects).
   alg :: sig m a -> m a
 
+instance Algebra Pure PureC where
+  alg v = case v of {}
+  {-# INLINE alg #-}
```
robrix: Defining the `Algebra` for `PureC` in here means we don’t need a `.hs-boot` file for it any more.

src/Control/Effect/Catch/Internal.hs
```diff
@@ -0,0 +1,20 @@
+{-# LANGUAGE DeriveFunctor, ExistentialQuantification, StandaloneDeriving #-}
+module Control.Effect.Catch.Internal
+( Catch(..)
+) where
+
+import Control.Effect.Class
+
+-- | 'Catch' effects can be used alongside 'Control.Effect.Throw.Throw' to provide recoverable exceptions.
+--
+-- @since 1.0.0.0
```
robrix: I double-checked that haddock propagates these docs along to `Control.Effect.Catch`.

src/Control/Effect/Catch/Internal.hs
```diff
@@ -0,0 +1,20 @@
+{-# LANGUAGE DeriveFunctor, ExistentialQuantification, StandaloneDeriving #-}
+module Control.Effect.Catch.Internal
```
robrix: I had initially placed all of these in a single `Control.Effect.Internal` module, but giving them their own modules means that much less recompilation.

robrix:
Oh dang it we merged this into `alg`, not `master` 😭

PR #300, opened 2019-10-24
Rename handle to thread

robrix:
This PR:

- [x] Depends on #295.
- [ ] Depends on #296.
- [x] Depends on #304.
- [x] Renames `handle` to `thread`.
- [x] Introduces a new `handle` function in the style of `handleIdentity`/`handleCoercible`, thus reserving the use of the word “handle” for things which eliminate effects.
- [x] Uses the new `handle` to tidy up a large number of algebras.

PR #303, opened 2019-10-25
Experiment: splitting up interpretation of the head and tail of the signature

robrix:
This PR is an experiment with doing something `MonadTransControl`-esque to deal with the tail of the signature and simplifying `Algebra` for the head.

- [x] Depends on #295.
- [ ] Depends on #296.
- [ ] Depends on #300.
- [x] Depends on #304.


robrix:
This is kind of cute, but entails some headaches, and ultimately isn’t worth the ergonomic problems.

PR #304, opened 2019-10-25
Acyclic imports, only this time merged into master

robrix:
Chained PRs are so easy to make mistakes with 😭

This is basically #299, only correctly using `master` as its base branch.


robrix:
#299 was approved so I’mma merge as soon as it goes green.

PR #305, opened 2019-10-26
Experiment: transformers & monad-control

robrix:
Experiment in standardizing on `transformers` & `monad-control` types & instances in place of our own definitions where feasible.

- [ ] Depends on #296.
- [x] 🔥 `PureC`.
- [x] 🔥 `ReaderC`.
- [x] 🔥 `StateC`.
- [ ] 🔥 `WriterC`.
- [ ] Use `MonadTransControl` to lift through `handle`.
- [ ] Use `MonadBaseControl` instead of `MonadUnliftIO`.


robrix:
Ended up getting a lot further with #306 & #307 than I thought I would, obsoleting most of what I was going for here.

PR #311, opened 2019-10-28
Simpler testing setup

robrix:
This PR 🔥s the dependency on `tasty-quickcheck` and thus on `QuickCheck`. This should speed up clean builds a bit (which are a bit more of a thing for the moment now that we’ve merged #195).


fused-effects.cabal
```diff
@@ -112,9 +112,11 @@ test-suite examples
   build-depends:
       base
     , fused-effects
+    , hedgehog           ^>= 1
+    , hedgehog-fn        ^>= 1
     , tasty              ^>= 1.2
+    , tasty-hedgehog     ^>= 1
```
robrix: this package’s name always makes me uncomfortable 🦔

PR #312, opened 2019-10-28
Script the release checklist

robrix:
1. We have a release checklist.

2. But checklists are _terrible_.

3. [Do-nothing scripts](https://blog.danslimmon.com/2019/07/15/do-nothing-scripting-the-key-to-gradual-automation/) are a much better alternative.

4. But I hate `bash`.

5. And I’ve been meaning to learn more about `runghc`.

6. And why not demonstrate using `fused-effects` to write multiple interpreters to do something practical, like a dry run mode?

Thus… this.


robrix:
Closing this. Have a better approach to the implementation of same in mind.

PR #318, opened 2019-10-30
Fail failed

robrix:
Turns out we were never actually _running_ the tests for `Fail`.

### DIFF
--- a/test/Catch.hs
+++ b/test/Catch.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE RankNTypes, ScopedTypeVariables, TypeApplications #-}
 module Catch
-( genN
+( tests
+, genN
 , test
 ) where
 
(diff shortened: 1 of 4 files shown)


test/Catch.hs
```diff
@@ -9,6 +10,11 @@ import Gen
 import Test.Tasty
 import Test.Tasty.Hedgehog
 
+tests :: TestTree
+tests = testGroup "Catch"
+  []
```
robrix: I stubbed this in so that it’s easier to see that all of the effect test modules are being run in `Test.hs`.

test/Test.hs
```diff
@@ -21,12 +23,14 @@ import Test.Tasty
...
 main = defaultMain $ testGroup "unit tests"
-  [ Choose.tests
+  [ Catch.tests
+  , Choose.tests
   , Cull.tests
   , Cut.tests
   , Cut.Church.tests
   , Empty.tests
   , Error.tests
+  , Fail.tests
```
robrix: 🤦‍♂

PR #321, opened 2019-11-01
Attempt to use preview of new actions cache

tclem:
This will likely require some tweaking to get right, but wanted to test out the preview of [caching](https://github.com/actions/cache) in GitHub Actions.

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -25,17 +25,17 @@ jobs:
       name: Cache ~/.cabal/packages
       with:
         path: ~/.cabal/packages
-        key: ${{ runner.os }}-${{ matrix.ghc }}-cabal-packages-${{ hashFiles('**/fused-effects.cabal') }}
+        key: ${{ runner.os }}-${{ matrix.ghc }}-cabal-packages
     - uses: actions/cache@preview
       name: Cache ~/.cabal/store
       with:
         path: ~/.cabal/store
-        key: ${{ runner.os }}-${{ matrix.ghc }}-cabal-store-${{ hashFiles('**/fused-effects.cabal') }}
+        key: ${{ runner.os }}-${{ matrix.ghc }}-cabal-store
     - uses: actions/cache@preview
       name: Cache dist-newstyle
       with:
         path: dist-newstyle
-        key: ${{ runner.os }}-${{ matrix.ghc }}-fused-effects-dist-${{ hashFiles('**/fused-effects.cabal') }}
+        key: ${{ runner.os }}-${{ matrix.ghc }}-fused-effects-dist
 
     - name: Install dependencies
       run: |
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -24,15 +24,15 @@ jobs:
     - uses: actions/cache@preview
       with:
         path: $HOME/.cabal/packages
-        key: ${{ runner.os }}-cabal-packages-${{ hashFiles(fused-effects.cabal) }}
+        key: ${{ runner.os }}-cabal-packages-${{ hashFiles('fused-effects.cabal') }}
     - uses: actions/cache@preview
       with:
         path: $HOME/.cabal/store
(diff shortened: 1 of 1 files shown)


.github/workflows/haskell.yml
```diff
@@ -21,6 +21,22 @@ jobs:
         ghc-version: ${{ matrix.ghc }}
         cabal-version: ${{ matrix.cabal }}
 
+    - uses: actions/cache@preview
+      name: Cache ~/.cabal/packages
+      with:
+        path: ~/.cabal/packages
+        key: ${{ runner.os }}-${{ matrix.ghc }}-cabal-packages-${{ hashFiles('**/fused-effects.cabal') }}
```
robrix: Should we exclude e.g. the hackage index since we regenerate it every time with `cabal update`?

.github/workflows/haskell.yml
```diff
@@ -21,6 +21,22 @@ jobs:
         ghc-version: ${{ matrix.ghc }}
```
robrix: Are you seeing a check warning on the first line? Any idea what it indicates?

.github/workflows/haskell.yml
```diff
@@ -21,6 +21,22 @@ jobs:
...
+    - uses: actions/cache@preview
+      name: Cache ~/.cabal/store
+      with:
+        path: ~/.cabal/store
+        key: ${{ runner.os }}-${{ matrix.ghc }}-cabal-store-${{ hashFiles('**/fused-effects.cabal') }}
+    - uses: actions/cache@preview
+      name: Cache dist-newstyle
+      with:
+        path: dist-newstyle
+        key: ${{ runner.os }}-${{ matrix.ghc }}-fused-effects-dist-${{ hashFiles('**/fused-effects.cabal') }}
```
robrix: What’s the significance of including the hash of the cabal file in the key? I get why we’d want to include the OS & ghc version, but I think the cabal file shouldn’t invalidate the cache, but rather update it.

tclem: We can take this out for sure. Most of the [examples](https://github.com/actions/cache/blob/master/examples.md) show basing the key off of some sort of lock file and invalidating the cache when dependencies change.

.github/workflows/haskell.yml
tclem: I do in early check runs, but not on the later ones?

.github/workflows/haskell.yml
tclem: Yes! Do you know where that file is, and does it exist before we run this cache action?

robrix: These are the files that we remove with travis in `semantic` (and which we got from the haskell-ci project):

https://github.com/github/semantic/blob/master/.travis.yml#L10-L18

I _think_ they shouldn’t exist until we’ve done a `cabal update` (unless they were restored there via an earlier cache).

.github/workflows/haskell.yml
robrix: Like so?

Screen Shot 2019-11-04 at 11 00 29 AM

Hopefully it’s benign…

.github/workflows/haskell.yml
robrix: I guess we could do a `cabal freeze` in CI, hash that, and then delete the freeze file. What do you think?

.github/workflows/haskell.yml
tclem: I'm going to leave this alone for now, we might cache stuff that gets overridden but I think that's ok. In the future, if we want to do this I think it'll be another step like this:

```
    - name: Clear out hackage index
      run: |
        rm -fv $HOME/.cabal/packages/hackage.haskell.org/00-index.*
        etc...
```

.github/workflows/haskell.yml
tclem: Just removed the hash of the cabal file entirely in e0479d86bbb213e9852cfd0f0b647d7a0aedd3aa, let's see what it does!

PR #359, opened 2020-03-02
Derivable Algebra

robrix:
This PR changes `Algebra`’s `alg` method, giving it a parameter for a monad homomorphism which is applied to actions within the signature.

Some neat effects of this:

- We can now derive `Algebra` instances using `GeneralizedNewtypeDeriving` or `DerivingVia`.
- The homomorphism obviates the need for uses of `hmap` in the case of e.g. `ReaderC`.
- The homomorphism obviates the need for uses of `handleCoercible` in the case of non-orthogonal carriers.
- All together now: the homomorphism obviates the need for `HFunctor` altogether.

There’s a bunch of work to do:

- [x] Depends on #354 and a hackage release containing it.
- [x] ~~Document the homomorphism parameter. Why is it a homomorphism? What does that mean? What should callers pass? (Why are you calling `alg` yourself, anyway?) What should instances do with it?~~ Not going to bother with this in this PR since #361 replaces it anyway.
- [x] 🔥 `handleCoercible`.
- [x] 🔥 `HFunctor` and its generic derivation machinery.

### DIFF
--- a/src/Control/Algebra.hs
+++ b/src/Control/Algebra.hs
@@ -164,6 +164,11 @@ instance Algebra sig m => Algebra (Reader r :+: sig) (Reader.ReaderT r m) where
   alg (L (Local f m k)) = Reader.local f m >>= k
   alg (R other)         = Reader.ReaderT $ \ r -> alg (hmap (`Reader.runReaderT` r) other)
 
+instance Algebra' sig m => Algebra' (Reader r :+: sig) (Reader.ReaderT r m) where
+  alg' (L (Ask       k)) = Reader.ask >>= k
+  alg' (L (Local f m k)) = Reader.local f m >>= k
+  alg' (R other)         = Reader.ReaderT $ \ r -> alg' (hmap (`Reader.runReaderT` r) other)
+
 newtype RWSTF w s a = RWSTF { unRWSTF :: (a, s, w) }
   deriving (Functor)
 
--- a/src/Control/Algebra.hs
+++ b/src/Control/Algebra.hs
@@ -127,6 +127,11 @@ instance (Algebra sig m, Effect sig) => Algebra (Error e :+: sig) (Except.Except
   alg (L (R (Catch m h k))) = Except.catchE m h >>= k
   alg (R other)             = Except.ExceptT $ alg (thread (Right ()) (either (pure . Left) Except.runExceptT) other)
 
+instance (Algebra' sig m, Effect sig) => Algebra' (Error e :+: sig) (Except.ExceptT e m) where
+  alg' _   (L (L (Throw e)))     = Except.throwE e
+  alg' hom (L (R (Catch m h k))) = Except.catchE (hom m) (hom . h) >>= hom . k
+  alg' hom (R other)             = Except.ExceptT $ alg' id (thread (Right ()) (either (pure . Left) (Except.runExceptT . hom)) other)
+
 instance Algebra sig m => Algebra sig (Identity.IdentityT m) where
   alg = Identity.IdentityT . alg . handleCoercible
 
--- a/src/Control/Algebra.hs
+++ b/src/Control/Algebra.hs
@@ -182,7 +182,7 @@ instance Algebra sig m => Algebra (Reader r :+: sig) (Reader.ReaderT r m) where
 instance Algebra' sig m => Algebra' (Reader r :+: sig) (Reader.ReaderT r m) where
   alg' hom (L (Ask       k)) = Reader.ask >>= hom . k
   alg' hom (L (Local f m k)) = Reader.local f (hom m) >>= hom . k
-  alg' hom (R other)         = Reader.ReaderT $ \ r -> alg' id (hmap ((`Reader.runReaderT` r) . hom) other)
+  alg' hom (R other)         = Reader.ReaderT $ \ r -> alg' ((`Reader.runReaderT` r) . hom) other
 
 newtype RWSTF w s a = RWSTF { unRWSTF :: (a, s, w) }
   deriving (Functor)
--- a/src/Control/Algebra.hs
+++ b/src/Control/Algebra.hs
@@ -15,6 +15,7 @@ An instance of the 'Algebra' class defines an interpretation of an effect signat
 -}
 module Control.Algebra
 ( Algebra(..)
+, Algebra'(..)
 , run
 , Has
 , send
@@ -58,6 +59,10 @@ import           Data.Tuple (swap)
 class (HFunctor sig, Monad m) => Algebra sig m | m -> sig where
   -- | Construct a value in the carrier for an effect signature (typically a sum of a handled effect and any remaining effects).
   alg :: sig m a -> m a
+
+
+class (HFunctor sig, Monad m) => Algebra' sig m | m -> sig where
+  alg' :: sig m a -> m a
 
 
 -- | Run an action exhausted of effects to produce its final result value.
--- a/src/Control/Algebra.hs
+++ b/src/Control/Algebra.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE FunctionalDependencies #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeOperators #-}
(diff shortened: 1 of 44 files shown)


src/Control/Algebra.hs
```diff
@@ -55,9 +58,9 @@ import           Data.Tuple (swap)
 -- | The class of carriers (results) for algebras (effect handlers) over signatures (effects), whose actions are given by the 'alg' method.
 --
 -- @since 1.0.0.0
-class (HFunctor sig, Monad m) => Algebra sig m | m -> sig where
+class Monad m => Algebra sig m | m -> sig where
   -- | Construct a value in the carrier for an effect signature (typically a sum of a handled effect and any remaining effects).
-  alg :: sig m a -> m a
+  alg :: Monad n => (forall x . n x -> m x) -> sig n a -> m a
```
robrix: This is the key change: `sig` now holds `n`s instead of `m`s, and a monad homomorphism is provided to map them into the carrier.

src/Control/Algebra.hs
```diff
@@ -79,51 +82,55 @@ type Has eff sig m = (Members eff sig, Algebra sig m)
 
 -- | Construct a request for an effect to be interpreted by some handler later on.
 send :: (Member eff sig, Algebra sig m) => eff m a -> m a
-send = alg . inj
+send = alg id . inj
```
robrix: `send` always operates in the topmost algebra, so we start by just passing `id`.

src/Control/Algebra.hs
```diff
@@ -79,51 +82,55 @@ type Has eff sig m = (Members eff sig, Algebra sig m)
...
-  alg (LiftWith with k) = with (Identity ()) coerce >>= k . runIdentity
+  alg hom (LiftWith with k) = with (Identity ()) (coerce . hom . runIdentity) >>= hom . k . runIdentity
 
 instance Algebra (Lift Identity) Identity where
-  alg (LiftWith with k) = with (Identity ()) coerce >>= k . runIdentity
+  alg hom (LiftWith with k) = with (Identity ()) (coerce . hom . runIdentity) >>= hom . k . runIdentity
 
 instance Algebra Choose NonEmpty where
-  alg (Choose m) = m True S.<> m False
+  alg hom (Choose m) = hom (m True) S.<> hom (m False)
```
robrix: Terminal algebras only need to apply the homomorphism to any continuations or higher-order positions to produce an action in their carrier.

src/Control/Algebra.hs
```diff
@@ -79,51 +82,55 @@ type Has eff sig m = (Members eff sig, Algebra sig m)
...
 -- transformers
 
 instance (Algebra sig m, Effect sig) => Algebra (Error e :+: sig) (Except.ExceptT e m) where
-  alg (L (L (Throw e)))     = Except.throwE e
-  alg (L (R (Catch m h k))) = Except.catchE m h >>= k
-  alg (R other)             = Except.ExceptT $ alg (thread (Right ()) (either (pure . Left) Except.runExceptT) other)
+  alg hom = \case
+    L (L (Throw e))     -> Except.throwE e
+    L (R (Catch m h k)) -> Except.catchE (hom m) (hom . h) >>= hom . k
+    R other             -> Except.ExceptT $ alg id (thread (Right ()) (either (pure . Left) (Except.runExceptT . hom)) other)
```
robrix: Nonterminal algebras, on the other hand, also need to apply the homomorphism before threading through the rest of the signature.

src/Control/Algebra.hs
```diff
@@ -151,13 +157,13 @@ instance Algebra sig m => Algebra sig (Ap m) where
...
+deriving instance Algebra sig m => Algebra sig (Alt m)
 
 instance Algebra sig m => Algebra (Reader r :+: sig) (Reader.ReaderT r m) where
-  alg (L (Ask       k)) = Reader.ask >>= k
-  alg (L (Local f m k)) = Reader.local f m >>= k
-  alg (R other)         = Reader.ReaderT $ \ r -> alg (hmap (`Reader.runReaderT` r) other)
+  alg hom = \case
+    L (Ask       k) -> Reader.ask >>= hom . k
+    L (Local f m k) -> Reader.local f (hom m) >>= hom . k
+    R other         -> Reader.ReaderT $ \ r -> alg ((`Reader.runReaderT` r) . hom) other
```
robrix: Algebras using `hmap` or `handleCoercible` can instead compose their handler onto the homomorphism passed to `alg` for the rest of the stack.

src/Control/Algebra.hs
```diff
@@ -167,43 +173,49 @@ toRWSTF w (a, s, w') = RWSTF (a, s, mappend w w')
...
-  alg (R (R (R other)))      = RWS.Lazy.RWST $ \ r s -> unRWSTF <$> alg (thread (RWSTF ((), s, mempty)) (\ (RWSTF (x, s, w)) -> toRWSTF w <$> RWS.Lazy.runRWST x r s) other)
+  alg hom = \case
+    L (Ask       k)      -> RWS.Lazy.ask >>= hom . k
+    L (Local f m k)      -> RWS.Lazy.local f (hom m) >>= hom . k
+    R (L (Tell w k))     -> RWS.Lazy.tell w *> hom k
+    R (L (Listen m k))   -> RWS.Lazy.listen (hom m) >>= hom . uncurry (flip k)
+    R (L (Censor f m k)) -> RWS.Lazy.censor f (hom m) >>= hom . k
+    R (R (L (Get   k)))  -> RWS.Lazy.get >>= hom . k
+    R (R (L (Put s k)))  -> RWS.Lazy.put s *> hom k
+    R (R (R other))      -> RWS.Lazy.RWST $ \ r s -> unRWSTF <$> alg id (thread (RWSTF ((), s, mempty)) (\ (RWSTF (x, s, w)) -> toRWSTF w <$> RWS.Lazy.runRWST (hom x) r s) other)
```
robrix: I also took the opportunity to rewrite _all_ the algebras using `LambdaCase`, because

1. I find `case` easier to read than all the chaff on the left, and

2. `alg hom x = case x of …` is the quintessential case for `\case`.

src/Control/Carrier/Choose/Church.hs
```diff
@@ -89,8 +90,10 @@ instance MonadTrans ChooseC where
...
-  alg (L (Choose k)) = ChooseC $ \ fork leaf -> fork (runChoose fork leaf (k True)) (runChoose fork leaf (k False))
-  alg (R other)      = ChooseC $ \ fork leaf -> alg (thread (pure ()) dst other) >>= runIdentity . runChoose (coerce fork) (coerce leaf) where
-    dst :: Applicative m => ChooseC Identity (ChooseC m a) -> m (ChooseC Identity a)
-    dst = runIdentity . runChoose (liftA2 (liftA2 (<|>))) (pure . runChoose (liftA2 (<|>)) (pure . pure))
+  alg (hom :: forall x . n x -> ChooseC m x) = \case
+    L (Choose k) -> ChooseC $ \ fork leaf -> fork (runChoose fork leaf (hom (k True))) (runChoose fork leaf (hom (k False)))
+    R other      -> ChooseC $ \ fork leaf -> alg id (thread (pure ()) dst other) >>= runIdentity . runChoose (coerce fork) (coerce leaf)
+    where
+    dst :: ChooseC Identity (n a) -> m (ChooseC Identity a)
+    dst = runIdentity . runChoose (liftA2 (liftA2 (<|>))) (pure . runChoose (liftA2 (<|>)) (pure . pure) . hom)
```
robrix: We use `ScopedTypeVariables` now since we close over `hom`. We could probably have removed the signature for `dst` instead, but the signatures for these are a bit subtle since they use the same carrier type but containing `Identity`, so I wanted to leave the signatures intact.

src/Control/Carrier/Cull/Church.hs
```diff
@@ -50,7 +51,7 @@ runCullM :: (Applicative m, Monoid b) => (a -> b) -> CullC m a -> m b
 runCullM leaf = runCull (liftA2 mappend) (pure . leaf) (pure mempty)
 
 -- | @since 1.0.0.0
-newtype CullC m a = CullC (ReaderC Bool (NonDetC m) a)
+newtype CullC m a = CullC { runCullC :: ReaderC Bool (NonDetC m) a }
```
robrix: I introduced selectors for a few types like this, but I don’t export them. I could maybe have used `coerce . hom` in the algebra instead of `runCullC . hom`, but it should be eliminated either way, and this way is clearer.

src/Control/Carrier/Error/Either.hs
```diff
@@ -43,7 +43,7 @@ runError (ErrorC m) = runExceptT m
 
 -- | @since 0.1.0.0
 newtype ErrorC e m a = ErrorC (ExceptT e m a)
-  deriving (Applicative, Functor, Monad, Fail.MonadFail, MonadFix, MonadIO, MonadTrans)
+  deriving (Algebra (Error e :+: sig), Applicative, Functor, Monad, Fail.MonadFail, MonadFix, MonadIO, MonadTrans)
```
robrix: This is one of the only cases where we define a carrier wrapping another one just so that we can customize some instance _other_ than `Algebra`, and so is likewise one of a small number of cases where we can derive an `Algebra` instance.

src/Control/Carrier/Interpret.hs
```diff
@@ -37,7 +38,7 @@ import           Unsafe.Coerce (unsafeCoerce)
 
 -- | A @Handler@ is a function that interprets effects described by @sig@ into the carrier monad @m@.
 newtype Handler sig m = Handler
-  { runHandler :: forall s x . sig (InterpretC s sig m) x -> InterpretC s sig m x }
+  { runHandler :: forall s n x . Monad n => (forall y . n y -> InterpretC s sig m y) -> sig n x -> InterpretC s sig m x }
```
robrix: This has to take a homomorphism just like `alg`, or else we have to keep `HFunctor` around. I think this is reasonable: the whole idea of `runInterpret` is that it allows you to write “a real algebra” inline.

src/Control/Effect/Class.hs
```diff
@@ -6,46 +6,18 @@
...
-  {-# INLINE hmap #-}
-
-
--- | Thread a 'Coercible' carrier through an 'HFunctor'.
---
---   This is applicable whenever @f@ is 'Coercible' to @g@, e.g. simple @newtype@s.
---
--- @since 1.0.0.0
-handleCoercible :: (HFunctor sig, Functor f, Coercible f g) => sig f a -> sig g a
-handleCoercible = hmap coerce
```
robrix: _All_ uses of `hmap` in the library are subsumed by the homomorphism parameter to `alg`, so we can eliminate `handleCoercible` along with the rest.

PR #361, opened 2020-03-03
Distributive Algebra

### LINKED ISSUE #289: Request for docs: Manual implementation of Effect instances
I've got a kind of hairy wrapper around some database query DSLs, and the `Effect` class is not automatically derivable due to some existentials. The docs don't really give me a good intuition around how the threading of effects is supposed to be done correctly. It might help to elaborate a bit on what the correct laws are for threading?

Minimal example to illustrate my confusion:

``` haskell
data RunDB backend m k
  = forall a . AttachName Text (backend -> m a) (a -> m k) 
  -- ^ passing a through from inner function 
  | forall a r . SqlSelect a r => Select_ (SqlQuery a) ([r] -> m k)
  -- ^ Weird one, [r] originates from an existentially quantified class
  --  / callback function type sig
  | Delete_ (SqlQuery ()) (m k)
  -- ^ Does the action that is the second argument need to do anything with state?
  | DeleteCount_ (SqlQuery ()) (Int64 -> m k) 
  -- ^ How about now
```

I'm hoping the when, why, & how to use the `f ()` argument of `handler` can be elucidated just a bit. Hope that makes sense!

confuse

robrix:
Building on #359, this PR replaces the monad homomorphism passed to `alg` with an initial state and a well-behaved distributive law, obviating the need for the `Effect` class altogether.

- [x] Depends on #354.
- [x] Depends on #359.
- [x] 🔥s `Effect`.
- [x] Document the new parameters. Why a distributive law? What does that mean? What should callers pass? (Why are you calling `alg` yourself, anyway?) What should instances do with it?
- [x] Fixes #289 (by removing `Effect`).


src/Control/Algebra.hs
```diff
@@ -59,8 +59,46 @@ import           Data.Tuple (swap)
...
+  --
+  -- By contrast, the 'Throw' case can simply return a value in 'Left', since there is no continuation to call—it represents an exceptional return—and @'Left' e :: forall a . Either e a@ (i.e. 'Left' is polymorphic in @a@).
+  --
+  -- Instances for monad transformers will most likely handle a signature containing multiple effects, with the tail of the signature handled by whatever monad the transformer wraps. In these cases, the tail of the signature can be delegated most conveniently using 'thread'; see the 'Algebra' instances for @transformers@ types such as 'Reader.ReaderT' and 'Except.ExceptT' for details.
+  alg
+    :: Functor ctx
+    => Handler ctx n m -- ^ A 'Handler' lowering computations inside the effect into the carrier type @m@.
+    -> sig n a         -- ^ The effect signature to be interpreted.
+    -> ctx ()          -- ^ The initial state.
+    -> m (ctx a)       -- ^ The interpretation of the effect in @m@.
```
robrix: Relative to `thread` and #359, this signature is somewhat different:

1. We use the `Handler` type synonym to make it clear what the purpose of the distributive law is.

2. We take the initial context in the final position, after the signature. This makes it clear that this is an algebra from `sig n a` to `ctx () -> m (ctx a)`, remembering that the handler reduces (context-borne) actions in `n` to (context-bearing) actions in `m`, i.e. we have what is essentially `sig (ctx ~> m :.: ctx) ~> ctx ~> m :.: ctx)`.

src/Control/Algebra.hs
```diff
@@ -78,57 +116,61 @@ run = runIdentity
...
-    R (Catch m h k) -> either (hom . k <=< hom . h) (hom . k) (hom m)
+    R (Catch m h k) -> either (hdl . (<$ ctx) . h) pure (hdl (m <$ ctx)) >>= hdl . fmap k
 
 instance Algebra (Reader r) ((->) r) where
-  alg hom = \case
-    Ask       k -> \ r -> hom (k r) r
-    Local f m k -> \ r -> hom (k (hom m (f r))) r
+  alg hdl sig ctx = case sig of
+    Ask       k -> \ r -> hdl (k r <$ ctx) r
+    Local f m k -> \ r -> hdl (fmap k (hdl (m <$ ctx) (f r))) r
```
robrix: We lose `LambdaCase` with the new parameter ordering, but it could be recovered by writing this tacitly in the context functor, e.g. writing the initial action as a continuation from `ctx ()` and then composing continuations with `>=>`. I’ve opted not to do that here because it doesn’t make the instances any clearer.

src/Control/Carrier/Interpret.hs
```diff
@@ -36,9 +35,9 @@ import           Control.Monad.Trans.Class
 import           Data.Functor.Const (Const(..))
 import           Unsafe.Coerce (unsafeCoerce)
 
--- | A @Handler@ is a function that interprets effects described by @sig@ into the carrier monad @m@.
-newtype Handler sig m = Handler
-  { runHandler :: forall s n x . Monad n => (forall y . n y -> InterpretC s sig m y) -> sig n x -> InterpretC s sig m x }
+-- | An @Interpreter@ is a function that interprets effects described by @sig@ into the carrier monad @m@.
+newtype Interpreter sig m = Interpreter
+  { runInterpreter :: forall ctx n s x . Functor ctx => Handler ctx n (InterpretC s sig m) -> sig n x -> ctx () -> InterpretC s sig m (ctx x) }
```
robrix: This was never a handler (which is what we call functions like `runState`), but rather an interpreter which now _takes_ a handler.

src/Control/Carrier/Interpret.hs
```diff
@@ -63,24 +62,24 @@ reify a k = unsafeCoerce (Magic k) a
...
+runInterpret f m = reify (Interpreter (\ hdl sig -> InterpretC . f (runInterpretC . hdl) sig)) (go m) where
   go :: InterpretC s eff m x -> Const (m x) s
   go (InterpretC m) = Const m
 
 -- | Interpret an effect using a higher-order function with some state variable.
 --
 -- @since 1.0.0.0
 runInterpretState
-  :: (forall n x . Monad n => (forall y . n y -> StateC s m y) -> s -> eff n x -> m (s, x))
+  :: (forall ctx n x . Functor ctx => Handler ctx n (StateC s m) -> eff n x -> s -> ctx () -> m (s, ctx x))
```
robrix: As with the context, the state `s` should follow the signature: `runInterpretState` interprets `eff` into (stateful) arrows modulo `ctx`.

src/Control/Carrier/NonDet/Church.hs
```diff
@@ -114,12 +113,12 @@ instance MonadTrans NonDetC where
...
     L (L Empty)      -> empty
-    L (R (Choose k)) -> hom (k True) <|> hom (k False)
-    R other          -> NonDetC $ \ fork leaf nil -> alg id (thread (pure ()) dst other) >>= runIdentity . runNonDet (coerce fork) (coerce leaf) (coerce nil)
+    L (R (Choose k)) -> hdl (k True <$ ctx) <|> hdl (k False <$ ctx)
+    R other          -> NonDetC $ \ fork leaf nil -> thread dst hdl other (pure ctx) >>= runIdentity . runNonDet (coerce fork) (coerce leaf) (coerce nil)
     where
-    dst :: NonDetC Identity (n a) -> m (NonDetC Identity a)
-    dst = runIdentity . runNonDet (liftA2 (liftA2 (<|>))) (Identity . runNonDetA . hom) (pure (pure empty))
+    dst :: NonDetC Identity (NonDetC m a) -> m (NonDetC Identity a)
+    dst = runIdentity . runNonDet (liftA2 (liftA2 (<|>))) (Identity . runNonDetA) (pure (pure empty))
```
robrix: Decomposing the distributive laws (by passing them separately to `thread`) allows you to understand both in isolation more easily.

src/Control/Effect/Catch/Internal.hs
```diff
@@ -5,15 +5,10 @@ module Control.Effect.Catch.Internal
...
 -- | 'Catch' effects can be used alongside 'Control.Effect.Throw.Throw' to provide recoverable exceptions.
 --
 -- @since 1.0.0.0
 data Catch e m k
   = forall b . Catch (m b) (e -> m b) (b -> m k)
 
 deriving instance Functor m => Functor (Catch e m)
-
-instance Effect (Catch e) where
-  thread ctx handler (Catch m h k) = Catch (handler (m <$ ctx)) (handler . (<$ ctx) . h) (handler . fmap k)
```
robrix: Not only do we no longer require the `Effect` instance, but we also don’t require the `Functor` instance. I intend to remove it in a future PR.

src/Control/Effect/Labelled.hs
```diff
@@ -39,22 +39,23 @@ import Control.Monad (MonadPlus)
...
 runLabelled :: forall label sub m a . Labelled label sub m a -> sub m a
 runLabelled (Labelled l) = l
 
 instance Algebra (eff :+: sig) (sub m) => Algebra (Labelled label eff :+: sig) (Labelled label sub m) where
-  alg hom = \case
-    L eff -> Labelled (alg (runLabelled . hom) (L (runLabelled eff)))
-    R sig -> Labelled (alg (runLabelled . hom) (R sig))
+  alg hdl = \case
+    L eff -> Labelled . alg (runLabelled . hdl) (L (runLabelled eff))
+    R sig -> Labelled . alg (runLabelled . hdl) (R sig)
```
robrix: Here’s a cute example of an algebra point-free in the context.

PR #365, opened 2020-03-12
Effects as GADTs

robrix:
This PR redefines all the effects as GADTs, simplifying and tidying the smart constructors and algebras considerably.

- [x] Depends on #361.
- [x] Update the changelog.

### DIFF
--- a/docs/defining_effects.md
+++ b/docs/defining_effects.md
@@ -5,7 +5,7 @@ Effects are a powerful mechanism for abstraction, and so defining new effects is
 It’s often helpful to start by specifying the types of the desired operations. For our example, we’re going to define a `Teletype` effect, with `read` and `write` operations, which read a string from some input and write a string to some output, respectively:
 
 ```haskell
-data Teletype m k
+data Teletype (m :: Type -> Type) k
 read :: Has Teletype sig m => m String
 write :: Has Teletype sig m => String -> m ()
 ```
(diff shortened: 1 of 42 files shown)


evanrelf:
Is this change mostly syntactic, or are there other implications for how effects are defined?

It looks like the GADT syntax will be a lot more straightforward and friendly compared to the old continuation style 🙂 I'm looking forward to it.

robrix:
> Is this change mostly syntactic, or are there other implications for how effects are defined?

Purely syntactic; it’s equivalent to the continuation style with a continuation of `pure`.

> It looks like the GADT syntax will be a lot more straightforward and friendly compared to the old continuation style 🙂 I'm looking forward to it.

Glad to hear it! I’m quite pleased with how it’s turning out.

docs/defining_effects.md
```diff
@@ -15,54 +15,45 @@ Effect types must have two type parameters: `m`, denoting any computations which
 Next, we can flesh out the definition of the `Teletype` effect by providing constructors for each primitive operation:
 
 ```haskell
-data Teletype m k
-  = Read (String -> m k)
-  | Write String (m k)
-  deriving (Functor)
+data Teletype (m :: Type -> Type) k where
+  Read  ::           Teletype m String
+  Write :: String -> Teletype m ()
```
robrix: One consequence of this change is that first-order effects in general require kind annotations for the `m` parameter. (This is documented in the common errors document.)

docs/defining_effects.md
```diff
@@ -15,54 +15,45 @@ Effect types must have two type parameters: `m`, denoting any computations which
...
 Now that we have our effect datatype, we can give definitions for `read` and `write`:
 
 ```haskell
 read :: Has Teletype sig m => m String
-read = send (Read pure)
+read = send Read
 
 write :: Has Teletype sig m => String -> m ()
-write s = send (Write s (pure ()))
+write s = send (Write s)
```
robrix: Having to provide continuations in smart constructors was a flaw that I am very glad to see removed.

docs/faqs.md
```diff
@@ -32,22 +43,37 @@ okay = do
...
+
+3. Use `Control.Effect.Labelled` to define an instance for some specific label:
+
+    ```haskell
+    instance HasLabelled State (State s) sig m => MTL.MonadState s (Wrapper s m) where
+      get = Control.Effect.State.Labelled.get @State
+      put = Control.Effect.State.Labelled.put @State
+    ```
+
+    Now `Wrapper` has a `MonadState` instance whenever `m` has an appropriately-labelled `State` effect, which can be provided by the `Control.Effect.Labelled.Labelled` carrier.
```
robrix: Figured I’d document using `Labelled` for `mtl` interop while I was in here.

docs/migrating.md
```diff
@@ -2,6 +2,7 @@
 
 This file summarizes the changelog and extracts the pieces most relevant to bringing up `fused-effects` applications to use newer versions of the library.
 
+
```
robrix: double 👏 space 👏 above 👏 headings

examples/Parser.hs
```diff
@@ -137,12 +140,12 @@ newtype ParseC m a = ParseC { runParseC :: StateC String m a }
 
 instance (Alternative m, Algebra sig m) => Algebra (Symbol :+: sig) (ParseC m) where
   alg hdl sig ctx = case sig of
-    L (Satisfy p k) -> do
+    L (Satisfy p) -> do
       input <- ParseC get
       case input of
-        c:cs | p c -> ParseC (put cs) *> hdl (k c <$ ctx)
+        c:cs | p c -> c <$ ctx <$ ParseC (put cs)
```
robrix: I really like the left-associativity of `<$` here ✨

src/Control/Algebra.hs
```diff
@@ -131,46 +131,46 @@ send sig = runIdentity <$> alg (fmap Identity . runIdentity) (inj sig) (Identity
...
+    Ask       -> (<$ ctx)
+    Local f m -> hdl (m <$ ctx) . f
 
 instance Algebra NonDet [] where
-  alg hdl sig ctx = case sig of
-    L Empty      -> []
-    R (Choose k) -> hdl (k True <$ ctx) ++ hdl (k False <$ ctx)
+  alg _ sig ctx = case sig of
+    L Empty  -> []
+    R Choose -> [ True <$ ctx, False <$ ctx ]
```
robrix: This is one of the best examples of how cool this change turned out to be. Lacking a continuation means we have to `return` a value instead of calling `k` with it, and that value is `ctx a` (whatever `a` is for that constructor). Since `Choose :: Choose m Bool`, that means we need to return `[ctx Bool]`. Hence, instead of concatenating the lists produced by the continuations, we produce the two-element list above and let the `Monad` instance for `[]` take care of the rest externally to `alg`.

All in all, this change should result in far fewer binds being issued. Many of these would ideally have been compiled away anyhow, but there’s now that much less work for the simplifier to do. And since the algebras themselves are (often dramatically!) simpler as a result, it’s win-win-win.

src/Control/Carrier/Choose/Church.hs
```diff
@@ -91,8 +92,8 @@ instance MonadTrans ChooseC where
 
 instance Algebra sig m => Algebra (Choose :+: sig) (ChooseC m) where
   alg (hdl :: forall x . ctx (n x) -> ChooseC m (ctx x)) sig (ctx :: ctx ()) = case sig of
-    L (Choose k) -> ChooseC $ \ fork leaf -> fork (runChoose fork leaf (hdl (k True <$ ctx))) (runChoose fork leaf (hdl (k False <$ ctx)))
-    R other      -> ChooseC $ \ fork leaf -> thread dst hdl other (pure ctx) >>= runIdentity . runChoose (coerce fork) (coerce leaf)
+    L Choose -> ChooseC $ \ fork leaf -> fork (leaf (True <$ ctx)) (leaf (False <$ ctx))
```
robrix: This is the Church-encoded binary tree version of that algebra for lists above.

src/Control/Effect/Fresh.hs
```diff
@@ -18,10 +18,11 @@ module Control.Effect.Fresh
 ) where
 
 import Control.Algebra
+import Data.Kind (Type)
 
 -- | @since 0.1.0.0
-newtype Fresh m k = Fresh (Int -> m k)
-  deriving (Functor)
+data Fresh (m :: Type -> Type) k where
```
robrix: I’ve gone out of my way to use `Type` instead of `*` for the new kind signatures. I’m planning on turning the star is type warning back on at some point this release.

PR #366, opened 2020-03-13
More transformers instances

robrix:
This PR adds `Algebra` instances for a few more types from `transformers`, namely including two only available in 0.5.6+.

- [x] Depends on #365.
- [x] Changelog.
- [x] Reference the new instances in the effect modules.
- [x] Tests.

### DIFF
--- a/src/Control/Algebra.hs
+++ b/src/Control/Algebra.hs
@@ -60,7 +60,6 @@ import           Data.Functor.Compose
 import           Data.Functor.Identity
 import           Data.List.NonEmpty (NonEmpty(..))
 import           Data.Monoid
-import           Data.Tuple (swap)
 
 -- | The class of carriers (results) for algebras (effect handlers) over signatures (effects), whose actions are given by the 'alg' method.
 --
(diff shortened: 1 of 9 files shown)


src/Control/Algebra.hs
```diff
@@ -201,6 +207,11 @@ deriving instance Algebra sig m => Algebra sig (Ap m)
 -- @since 1.0.1.0
 deriving instance Algebra sig m => Algebra sig (Alt m)
 
+instance Algebra sig m => Algebra (Empty :+: sig) (Maybe.MaybeT m) where
+  alg hdl sig ctx = case sig of
+    L Empty -> Maybe.MaybeT (pure Nothing)
+    R other -> Maybe.MaybeT $ thread (maybe (pure Nothing) Maybe.runMaybeT) hdl other (Just ctx)
+
```
robrix: I honestly don’t know why I didn’t implement this before.

src/Control/Algebra.hs
```diff
@@ -214,10 +225,26 @@ toRWSTF :: Monoid w => w -> (a, s, w) -> RWSTF w s a
 toRWSTF w (a, s, w') = RWSTF (a, s, mappend w w')
 {-# INLINE toRWSTF #-}
 
+newtype Swap s a = Swap { getSwap :: (a, s) }
+  deriving (Functor)
```
robrix: This is handy for giving us a `Functor` instance for the backwards pairs `transformers` is so fond of.

src/Control/Algebra.hs
```diff
@@ -244,24 +271,33 @@ instance Algebra sig m => Algebra (State s :+: sig) (State.Lazy.StateT s m) wher
...
+    R other   -> State.Strict.StateT $ \ s -> getSwap <$> thread (fmap Swap . uncurry State.Strict.runStateT . getSwap) hdl other (Swap (ctx, s))
+
+#if MIN_VERSION_transformers(0,5,6)
+instance (Algebra sig m, Monoid w) => Algebra (Writer w :+: sig) (Writer.CPS.WriterT w m) where
+  alg hdl sig ctx = case sig of
+    L (Tell w)     -> ctx <$ Writer.CPS.tell w
+    L (Listen m)   -> swapAndLift <$> Writer.CPS.listen (hdl (m <$ ctx))
+    L (Censor f m) -> Writer.CPS.censor f (hdl (m <$ ctx))
+    R other        -> Writer.CPS.writerT $ getSwap <$> thread (\ (Swap (x, s)) -> Swap . fmap (mappend s) <$> Writer.CPS.runWriterT x) hdl other (Swap (ctx, mempty))
+#endif
```
robrix: I had implemented this before, but removed it because it was unavailable on some of our `ghc`s. But since we added `CPP` to support the `Ap` instance I figured I might as well add it back in.

PR #368, opened 2020-03-15
Benchmark reorganization

robrix:
This PR moves the benchmarks around a little and simplifies them to not benchmark 100/1000/10000 iterations separately, but just run a large number of iterations once.

### DIFF
--- a/benchmark/Bench/Interpret.hs
+++ b/benchmark/Bench/Interpret.hs
@@ -9,7 +9,7 @@ module Bench.Interpret
 
 import Control.Carrier.Interpret
 import Control.Carrier.State.Strict
-import Control.Monad (replicateM_)
+import Data.Foldable (for_)
 import Gauge hiding (benchmark)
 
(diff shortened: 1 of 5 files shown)


benchmark/Bench/Interpret.hs
```diff
@@ -0,0 +1,47 @@
...
+        Get   -> gets @Int (<$ ctx)
+        Put s -> ctx <$ put s) $ modLoop n) n
+    , bench "InterpretStateC" $
+      whnf (\ n -> fst . run $ runInterpretState (\ _ (sig :: State Int m k) (s :: Int) ctx -> case sig of
+        Get   -> pure (s, s <$ ctx)
+        Put s -> pure (s, ctx)) 0 $ modLoop n) n
+    , bench "StateC" $
+      whnf (run . execState @Int 0 . modLoop) n
+    ]
+  , bgroup "IO"
```
robrix: I’m running the benchmarks both in `Identity` and in `IO` now.

PR #369, opened 2020-03-15
Church-encoded Writer carrier

robrix:
This PR introduces and benchmarks a high-performance Church-encoded `Writer` carrier. In benchmarks where it wraps `IO`, it outperforms every other carrier by quite a decent margin.

### DIFF
--- a/src/Control/Carrier/State/Lazy.hs
+++ b/src/Control/Carrier/State/Lazy.hs
@@ -99,9 +99,9 @@ instance Monad m => Applicative (StateC s m) where
   {-# INLINE (<*) #-}
 
 instance Monad m => Monad (StateC s m) where
-  m >>= k = StateC $ \ s -> do
-    ~(s', a) <- runState s m
-    runState s' (k a)
+  StateC a >>= k = StateC $ \ s -> do
+    ~(s', a') <- a s
+    runState s' (k a')
   {-# INLINE (>>=) #-}
 
 instance (Alternative m, Monad m) => Alternative (StateC s m) where
--- a/src/Control/Carrier/State/Lazy.hs
+++ b/src/Control/Carrier/State/Lazy.hs
@@ -23,7 +23,7 @@ module Control.Carrier.State.Lazy
 ) where
 
 import Control.Algebra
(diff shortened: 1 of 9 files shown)


src/Control/Carrier/State/Lazy.hs
```diff
@@ -71,7 +71,7 @@ execState s = fmap fst . runState s
 {-# INLINE[3] execState #-}
 
 -- | @since 1.0.0.0
-newtype StateC s m a = StateC { runStateC :: s -> m (s, a) }
+newtype StateC s m a = StateC (s -> m (s, a))
```
robrix: Whoops.

src/Control/Carrier/Writer/Strict.hs
```diff
@@ -60,15 +61,18 @@ newtype WriterC w m a = WriterC { runWriterC :: StateC w m a }
...
-    L (Censor f m) -> WriterC . StateC $ \ w -> do
-      (w', a) <- runWriter (hdl (m <$ ctx))
-      let w'' = mappend w (f w')
-      w'' `seq` pure (w'', a)
-    R other        -> WriterC (alg (runWriterC . hdl) (R other) ctx)
+  alg hdl sig ctx = WriterC $ case sig of
+    L writer -> StateC $ \ w -> case writer of
+      Tell w'    -> do
+        let !w'' = mappend w w'
+        pure (w'', ctx)
```
robrix: `mappend`ing strictly here made for a ~2.5x improvement to the performance of this carrier in the benchmarks.

src/Control/Carrier/Writer/Strict.hs
```diff
@@ -60,15 +61,18 @@ newtype WriterC w m a = WriterC { runWriterC :: StateC w m a }
...
+      Listen   m -> do
+        (w', a) <- runWriter (hdl (m <$ ctx))
+        let !w'' = mappend w w'
+        pure (w'', (,) w' <$> a)
```
robrix: I strongly prefer `BangPatterns` to `seq` for these; plus, it turns out that we were inadvertently making `listen` and `censor` _too_ strict.

PR #374, opened 2020-03-24
[WIP] Use Close type family to avoid ambiguity

zhujinxuan:
This PR try to introduce http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.643.3533&rep=rep1&type=pdf to eliminate ambiguity by abitary `:+:`.

The paper has an implementation by its author as in https://github.com/pa-ba/compdata

Currently, this PR is work in process to fix about Label.

### DIFF
--- a/fused-effects.cabal
+++ b/fused-effects.cabal
@@ -99,6 +99,7 @@ library
     Control.Effect.State.Internal
     Control.Effect.Throw.Internal
     Control.Effect.Writer.Internal
+    Control.Effect.Close
   build-depends:
     , base          >= 4.9 && < 4.14
     , transformers  >= 0.4 && < 0.6
--- a/src/Control/Effect/Close.hs
+++ b/src/Control/Effect/Close.hs
@@ -1,0 +1,63 @@
+{-# LANGUAGE DeriveTraversable #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE PolyKinds #-}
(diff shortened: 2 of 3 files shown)


robrix:
Interesting! Which ambiguity are you intending to resolve? I’m familiar with `compdata` and the limitations of _Data Types à la Carte_, but I’m curious what the practical impact for end users is going to be.

zhujinxuan:
What I want to resolve is things like 

```
alg :: s -> s :<+>:  t :<+>: s 
```

There are two possible definitions, by the leftmost s and right most s(Though we use overlappable in this library to cast it to the leftmost).  Though it may not hurt in practice, I think it is better to avoid such ambiguity.

This is another intention of this PR.  I am not sure whether `reassociateSumL` is the reason of https://github.com/fused-effects/fused-effects/issues/163.  If it is,  perhaps we can avoid it by `type Position` instead of `l :+: m :+: n` ->  `(l :+: m) :+: n`

zhujinxuan:
It seems too much change if we want to introduce Close Type Family like in compdata.  The haskell cannot inference the constraint well between current code and the constraint by the type family.  

I will close this PR this moment.  I may try to do some change about it in future.

robrix:
Ah, this is actually a feature of the library: inner carriers can shadow outer carriers, just like inner scopes can shadow outer scopes.

robrix:
Note also that `reassociateSumL` shouldn’t actually come into play in #163 at all, because it only applies when we nest sums on the left, which we only do for cases where we’ve got a type synonym for multiple effects which are handled together, like `Error` or `NonDet` (and which we haven’t seen slowdowns with).

PR #390, opened 2020-06-16
Fused effects 1.0.2.2 release

joshvera:
Cherry-picks https://github.com/fused-effects/fused-effects/pull/382 on top of 1.0.2.0 for 8.10.1 support.


evanrelf:
@joshvera It looks like every release of `fused-effects` so far has received a tag on the `master` branch, but that breaks down in this situation, since 1.0.2.2 is not based on the latest changes in `master`.

You might want to consider moving @robrix's unreleased changes to a separate branch, otherwise the purpose of the `master` branch remains unclear. Or you could make a branch for each release so that they aren't forced into a sequence. Either way, I think the project would benefit from a more explicit release workflow.

Hopefully that helps 🙂 That said, I'm just some random onlooker, so you can do whatever you want.

robrix:
@evanrelf we can tag based on this branch instead of based on `master`, that’s fine

PR #402, opened 2020-12-03
Derive MonadFix for labelled effects

JaSpa:
All other carriers already define a conditional `MonadFix` instance.

These instances should probably get a `@since XX` annotation and maybe an entry in the changelog. What is your process in that regard? I can add those if you can tell me what the next version number will be or alternatively add some kind of placeholder to fill in when the next release is done.

### DIFF
--- a/src/Control/Effect/Labelled.hs
+++ b/src/Control/Effect/Labelled.hs
@@ -37,6 +37,7 @@ import Control.Applicative (Alternative)
 import Control.Effect.Sum (reassociateSumL)
 import Control.Monad (MonadPlus)
 import Control.Monad.Fail as Fail
+import Control.Monad.Fix
 import Control.Monad.IO.Class
 import Control.Monad.Trans.Class
 import Data.Functor.Identity
@@ -46,7 +47,17 @@ import Data.Kind (Type)
 --
 -- @since 1.0.2.0
 newtype Labelled (label :: k) (sub :: (Type -> Type) -> (Type -> Type)) m a = Labelled (sub m a)
-  deriving (Alternative, Applicative, Functor, Monad, Fail.MonadFail, MonadIO, MonadPlus, MonadTrans)
+  deriving
+    ( Alternative
+    , Applicative
+    , Functor
+    , Monad
+    , Fail.MonadFail
+    , MonadFix -- ^ @since 1.1.1
+    , MonadIO
+    , MonadPlus
+    , MonadTrans
+    )
 
 -- | @since 1.0.2.0
 runLabelled :: forall label sub m a . Labelled label sub m a -> sub m a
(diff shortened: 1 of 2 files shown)


robrix:
For the CI failures, merging `master` should suffice.

There aren’t any other changes pending on `master`, so we’ll release this as 1.1.1. `@since 1.1.1` annotations and a changelog entry would be much appreciated, thank you!

src/Control/Effect/Labelled.hs
```diff
@@ -46,7 +47,7 @@ import Data.Kind (Type)
 --
 -- @since 1.0.2.0
 newtype Labelled (label :: k) (sub :: (Type -> Type) -> (Type -> Type)) m a = Labelled (sub m a)
-  deriving (Alternative, Applicative, Functor, Monad, Fail.MonadFail, MonadIO, MonadPlus, MonadTrans)
+  deriving (Alternative, Applicative, Functor, Monad, Fail.MonadFail, MonadIO, MonadPlus, MonadFix, MonadTrans)
```
robrix, line 50: Could you please keep this alphabetized (by class name)? i.e. insert `MonadFix` between `Fail.MonadFail` and `MonadIO`.

src/Control/Effect/Labelled.hs
```diff
@@ -114,7 +115,7 @@ sendLabelled op = runIdentity <$> alg (fmap Identity . runIdentity) (injLabelled
 --
 -- @since 1.0.2.0
 newtype UnderLabel (label :: k) (sub :: (Type -> Type) -> (Type -> Type)) (m :: Type -> Type) a = UnderLabel (m a)
-  deriving (Alternative, Applicative, Functor, Monad, Fail.MonadFail, MonadIO, MonadPlus)
+  deriving (Alternative, Applicative, Functor, Monad, Fail.MonadFail, MonadIO, MonadPlus, MonadFix)
```
robrix, line 118: Ditto here.

robrix (changes requested):

JaSpa:
@robrix I added the `@since` annotations and a changelog entry.

src/Control/Effect/Labelled.hs
```diff
@@ -46,7 +47,17 @@ import Data.Kind (Type)
...
+    ( Alternative
+    , Applicative
+    , Functor
+    , Monad
+    , Fail.MonadFail
+    , MonadFix -- ^ @since 1.1.1
+    , MonadIO
+    , MonadPlus
+    , MonadTrans
+    )
```
JaSpa, line 60: I think breaking this into multiple lines is cleaner than an inline comment `…, MonadFix {- ^ @since 1.1.1 -}, …`

robrix (approved):
❤️ perfect, thank you so much!

robrix:
[1.1.1](https://hackage.haskell.org/package/fused-effects-1.1.1.0) has been released!

PR #407, opened 2020-12-17
Fix possible typos in reinterpreting effects doc

shinzui:


docs/reinterpreting_effects.md
```diff
@@ -234,7 +234,7 @@ instance (MonadIO m, Algebra sig m) => Algebra (Http :+: sig) (MockHttpClient m)
     L (SendRequest req) -> do
       responder <- MockHttpClient ask
       (<$ ctx) <$> liftIO (responder req)
-    R other -> MockHttpClient (alg (runMockHttpClient . hdl) other ctx)
+    R other -> MockHttpClient (alg (runMockHttpClient . hdl) (R other) ctx)
```
ocharles, line 237: I usually use `inj` here - is there any preference to what we should do?

shinzui: I updated the doc to use `inj`. Is there any advantage of using it?

ocharles: Sorry, I wasn't trying to /suggest/ `inj`, just that I use it. I don't know which is better!

shinzui: I'm new to `fused-effects`, so I wouldn't know. But having to import `inj` from `Control.Effect.Sum` indicates, to me at least, that it's not meant to be used directly. Maybe @patrickt or  @robrix can weigh in.

robrix: Generally I use `R`, but `inj` is fine too. As @shinzui notes, `inj` requires an extra import, so I’d probably use `R` in the docs.

docs/reinterpreting_effects.md
```diff
@@ -181,7 +182,7 @@ handlePrint r =
     Left invalidContentTypeError -> print invalidContentTypeError
     Right ok -> case ok of
       Left jsonParseError -> print jsonParseError
-      Right facts -> traverse (putStrLn . catFact) facts
+      Right facts -> mapM_ (putStrLn . catFact) facts
```
robrix, line 185: Good catch 👍🏻

robrix (approved):
Nice, thank you! 👍🏻 I’m ok with leaving this using `inj`, or you can revert it to use `R`, at your option.

docs/reinterpreting_effects.md
ocharles: Cool, I'll probably switch to using `R` too then!

docs/reinterpreting_effects.md
jkachmar: Super tiny nitpick, but should be `mapM_` (in the spirit of this all being `Monad`ic anyway) or should it be `traverse_` (in the spirit of the original `traverse` call)?

shinzui: Switched to `traverse_`

shinzui:
I reverted back to using `R` and addressed the `mapM` comment. I can squash the commits if everything else looks fine.

docs/reinterpreting_effects.md
robrix: I’m good either way 👍🏻

robrix:
@shinzui I (very strongly) prefer non-squashed history. Thanks very much for the great PR!

PR #419, opened 2022-02-27
Instances for Choosing

robrix:
This PR defines/derives a number of instances for `Control.Effect.Choose.Choosing`.

These instances, much like the similar ones for `Alt` and `Ap`, allow us to build values in `Choosing` using effect operators directly, rather than requiring that the operators be wrapped in `Choosing` constructions manually.


patrickt (approved):

patrickt (changes requested):
Oh, wait, can you add a Changelog entry?

robrix:
@patrickt: Good catch. Changelog entry added.

patrickt (approved):

PR #421, opened 2022-03-01
Run CI against the latest point release of each major compiler release

robrix:
We were specifying point releases manually, which was a bit of a pain to maintain.


robrix:
Oh right, I ran into this before. Bummer.

PR #437, opened 2022-10-12
Use simple cabal cradle

fendor:
With cabal 3.4, this suffices for most (hopefully all) cabal projects.

Allows us to load the examples into the IDE.


robrix:
🤘🏻 nice one, thank you.

PR #439, opened 2022-11-24
Remove Deadpendency as a user

AlistairB:
As it is [shutting down](https://alistairb.dev/shutting-down-deadpendency/).

fused effects worked perfectly for me though! Oh well.


robrix:
:cry: (but ty!)

PR #453, opened 2024-03-23
Fix `Accum` carriers (`Strict` and `Church`) to not duplicate state

### LINKED ISSUE #449: Accum effect mixed with other effects duplicates accumulated monoidal value
Here is a minimal example that demonstrates the problem:
```
import Control.Carrier.Accum.Strict
import Control.Carrier.Lift
import Data.Monoid (Sum (..))

addOne :: (Has (Accum (Sum Integer)) sig m, Has (Lift IO) sig m) => m ()
addOne = do
    add (Sum (1 :: Integer))
    sendIO $ putStrLn "hi"

main :: IO ()
main = do
    s <- runM . execAccum (Sum (0 :: Integer)) $ addOne
    print (getSum s)
```
Since we simply run `addOne` once, which adds 1 to the accumulated value (which we start at 0), this should print `1`.  However, it prints `2`.  If we add a second call to `sendIO $ putStrLn "hi"`, it prints `4` (after printing `hi` twice, of course).  It seems like every additional bind causes the accumulated state to be combined with itself.

* `Control.Carrier.Accum.Church` and `Control.Carrier.Accum.Strict` both seem to have this issue, whereas `Control.Carrier.Accum.IORef` works as expected.
* The problem only seems to manifest when there is another effect mixed in besides the `Accum` effect.  This perhaps suggests that the issue is in the `Algebra (Accum w :+: sig) (AccumC w m)` instance, though I don't understand what's going on there well enough to be sure.

byorgey:
Fixes #449.  Previously, the `Algebra` instances erroneously returned a copy of the input state when processing other effects.  Since these carriers assume that the returned state represents any *additional* output produced, this resulted in the state being duplicated.  Instead, when processing other effects (which necessarily have no `Accum` effects) we should return `mempty`.

Sorry it's taken me a while to get around to this.  I had grand plans at first of looking into writing some additional tests, as discussed in #449 , but I'm not sure whether I'll ever get around to that, and I figure it's better to have a fix with no tests than nothing at all.

### DIFF
--- a/src/Control/Carrier/Accum/Church.hs
+++ b/src/Control/Carrier/Accum/Church.hs
@@ -129,5 +129,5 @@ instance (Algebra sig m, Monoid w) => Algebra (Accum w :+: sig) (AccumC w m) whe
     L accum -> case accum of
       Add w' -> k w' ctx
       Look   -> k mempty $ w <$ ctx
-    R other  -> thread (uncurry (runAccum (curry pure)) ~<~ hdl) other (w, ctx) >>= uncurry k
+    R other  -> thread (uncurry (runAccum (curry pure)) ~<~ hdl) other (mempty, ctx) >>= uncurry k
   {-# INLINE alg #-}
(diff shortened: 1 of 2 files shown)


patrickt:
@byorgey Sorry for letting this marinate for so long! I’m hitting weird permissions errors from this repo that @robrix is fixing, but once that’s fixed I’m going to merge this in (this PR’s CI runs seem busted, so I’ll do it in a new PR.)

patrickt:
Fixed in #460.

